Issue #18, opened 2020-02-23
Fix Unicode and Hex Escape Parsing in String/Symbol/CLOB


almann:
As per #17, there appears to be an issue around escape processing for Unicode escapes (i.e. `\uHHHH`, `\UHHHHHHHH`, and `\xHH`), we should make sure this is working correctly in parser, because for at least CLOBs, `\xHH` does not appear to process correctly when parsing `clobNewlines.ion` in the test vectors.

almann:
Looking at this more deeply, the part of the parsing that controls this is in the `String` struct (see [types_character.go#L31-L46][1]):

```go
	// Unquote translates escape sequences, e.g. `\u00ff`, into their utf8
	// normalized value, e.g. "ÿ".  We need to surround the text with double
	// quotes so that Unquote knows whether to treat it as literal text or not.
	var err error
	s.unquoted, err = strconv.Unquote(`"` + string(s.text) + `"`)
	if err != nil {
		s.unquoted = string(s.text)
	}
```

It is probably mostly working (needs stronger tests to verify this behavior), but the escape processing probably should not be spread across the lexer, parser, and the `String` element/value type.

[1]: https://github.com/amzn/ion-go/blob/11e5dc792d76ab8a940fe5c923f814f54c9a43bf/ion/types_character.go#L31-L46

Issue #17, opened 2020-02-23
String/Symbol/CLOB Literal Escape Processing


almann:
While investigating #3.  It seems that escapes pass through verbatim in text parsing; they are *recognized* by the lexer, but not replaced with the runes (code points) they represent.

E.g. adding the following to `lex_test.go`:

```go
		{
			name:     "quoted string with escapes",
			input:    []byte(`"\x41"`),
			expected: []Item{doubleQuote("A"), tEOF},
		},
```

Fails as follows:

```
    --- FAIL: TestLex/quoted_string_with_escapes (0.00s)
        lex_test.go:738: Expected: [<A> EOF]
        lex_test.go:739: Found:    [<\x41> EOF]
        lex_test.go:740: (-expected, +found)   []lex.Item{
            - 	s"<A>",
            + 	s`<\x41>`,
              	{Type: s"EOF"},
              }
```

Either I misunderstand the responsibility of the lexer, or this is a bug and we should make sure escapes are being processed correctly.

almann:
[parse_text_simple.go#L65-L170][1] shows where the processing of escapes are happening.  It duplicates the lexing code a bit, since the lexing code has to effectively parse the escapes when doing the normalization.  That said, I don't see this processing all the escapes (particularly `\u`, `\U`, and `\x`), so I think we need to take a closer look at this logic and its factoring.

[1]: https://github.com/amzn/ion-go/blob/11e5dc792d76ab8a940fe5c923f814f54c9a43bf/ion/parse_text_simple.go#L65-L170

PR #19, opened 2020-02-23
Fixes CLOB newline processing.

### LINKED ISSUE #3: Fix CLOB Newline Normalization
https://github.com/amzn/ion-tests/commit/2a1f3f4f272256108e89f1c53bfae5b3a9ede812 introduces CLOB newline normalization in text.  This is currently skipped in the parser tests.

almann:
Also adds escape processing in CLOB parsing to support `\xHH` escapes
which is part of #18.

#17 asks a larger question about the factoring around escape logic
in the parser, which this commit does not aim to address.

Fixes #3

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/ion/parse_text_simple.go
+++ b/ion/parse_text_simple.go
@@ -17,9 +17,9 @@ package ion
 
 import (
 	"bytes"
-	"math"
-
 	"github.com/amzn/ion-go/internal/lex"
+	"math"
+	"strconv"
 )
 
 // This file contains text parsers for Null, Padding, Bool, Symbol, String, Blob, and Clob.
@@ -115,7 +115,7 @@ func doStringReplacements(str []byte) []byte {
 	return ret
 }
 
-// doClobReplacements is like doStringReplacements, but keeps \r characters as-is.
+// doClobReplacements is like doStringReplacements but is restricted to escapes that CLOBs have
 func doClobReplacements(str []byte) []byte {
 	strLen := len(str)
 	ret := make([]byte, 0, strLen)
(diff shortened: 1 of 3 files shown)


ion/parse_text_simple.go
```diff
@@ -115,7 +115,7 @@ func doStringReplacements(str []byte) []byte {
 	return ret
 }
 
-// doClobReplacements is like doStringReplacements, but keeps \r characters as-is.
+// doClobReplacements is like doStringReplacements but is restricted to escapes that CLOBs have
```
mrutter-amzn: Nitpick: period

ion/parse_text_simple.go
```diff
@@ -124,11 +124,9 @@ func doClobReplacements(str []byte) []byte {
 		case '\r':
 			// We need to treat "\r\n" as "\n", so skip extra if what comes next is "\n".
```
mrutter-amzn: Update the comment since we are also normalizing `\r` to `\n`.

It's strange, this code is like this because there was a test that was failing if I didn't retain `\r`.  *shrug*.

ion/parse_text_simple.go
```diff
@@ -155,6 +153,12 @@ func doClobReplacements(str []byte) []byte {
 			case '\'', '"', '\\':
 				ret = append(ret, next)
 				index++
+			case 'x':
+				// Decode the hex sequence
+				index += 2
+				octet, _ := strconv.ParseUint(string(str[index:index+2]), 16, 8)
```
mrutter-amzn: Unprotected array index access after increment.  Also lacking handling if what is being escaped is an invalid hex number.

almann: It is technically an invariant from the lexer, would it be okay to panic in this case?

ion/parse_text_simple.go
almann: Fixed.

ion/parse_text_simple.go
almann: Fixed.

ion/parse_text_simple.go
almann: Actually, I checked and this is fixed in #20, so unless you object, I would defer to that PR.

mrutter-amzn: That's fine.  As to the general question - if it should be an invariant of the lexer then a panic with a message that includes contextual details would be warranted.

Issue #12, opened 2020-01-31
Reader (or Scanner)/Writer API


almann:
Had a a whiteboard session with @pbcornell and @mrutter-amzn on 1/24/2020 with some consensus on the need for lower level APIs around parsing and generating Ion data:

![20200124_113120](https://user-images.githubusercontent.com/503506/73559505-0ad03380-440a-11ea-9ff9-c704253828a8.jpg)

Spelling aside, I think this even oriented API is a very reasonable approach to defining the lower level APIs

PR #25, opened 2020-04-28
Reader Public API

matthewh-BQ:
Issue #12 

Work in progress. Initial review of Reader public API.

Some notes:

- EventType to handle start, end, and EOF. Currently using the container stack workflows, which will not satisfy the requirements. Though this is mostly implementation detail and shouldn't affect API.

- Binary and text readers should instantiate concrete implementation of binary and text events.

### DIFF
--- a/ion/catalog.go
+++ b/ion/catalog.go
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package ion
+
+type Catalog interface {
+	// Returns the symbol table with a specific name with the latest version.
+	LatestTable(name string) SymbolTable
+
+	// Returns a symbol table with a specific name and version.
+	ExactTable(name string, version int) SymbolTable
+}
--- a/ion/event.go
+++ b/ion/event.go
@@ -1,0 +1,227 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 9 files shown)


ion/catalog.go
```diff
@@ -0,0 +1,23 @@
...
+ * permissions and limitations under the License.
+ */
+package ion
+
+type Catalog interface {
+	// Returns the symbol table with a specific name with the latest version.
+	LatestTable(name string) SymbolTable
+
+	// Returns a symbol table with a specific name and version.
+	ExactTable(name string, version int) SymbolTable
```
therapon, line 22: If my understanding of Go naming is correct I  would name this `Table`

ion/event.go
```diff
@@ -0,0 +1,181 @@
...
+
+import (
+	"math/big"
+	"time"
+)
+
+type Event interface {
+	// Returns the depth into the Ion value that this Reader has traversed. At top level the depth is 0, and
+	// it increases by one on each call to StepIn().
+	Depth() int
```
therapon, line 26: Add comment for subtracting for each `StepOut()` we should try to be complete on documentation on interfaces

ion/event.go
```diff
@@ -0,0 +1,181 @@
...
+
+	// Returns the symbol table that is applicable to the current value.
+	SymbolTable() SymbolTable
+
+	// Returns the type of the current value, or null if there is no current value.
```
therapon, line 31: replace `null` with `nil`

ion/event.go
```diff
@@ -0,0 +1,181 @@
...
+
+	// Returns the current value's annotations as symbol tokens (text + ID).
+	TypeAnnotationSymbols() []SymbolToken
+
+	// Determines if the current value contains such annotation.
+	HasAnnotation(annotation string) bool
+
+	// Return the field name of the current value, or null if there is no current value.
+	FieldName() string
+
```
therapon, line 45: Given that the API here provides operations to test first if something is there (or the same type) I would expect the "getter" methods to always return with errors. I am not sure if this is the reccommended Go pattern though ...

ion/event.go
```diff
@@ -0,0 +1,181 @@
...
+	StringValue() (string, error)
+
+	// Returns the current value as a SymbolToken. It returns an error if the current value is not an Ion Symbol.
+	SymbolValue() (SymbolToken, error)
+
+	// Returns the current value as a byte slice. It returns an error if the current value is not an Ion Clob
+	// or an Ion Blob.
+	ByteValue() ([]byte, error)
+
+	EventType() container
```
therapon, line 92: I would expect `EventType` to be an enum here not an int

ion/reader.go
```diff
@@ -0,0 +1,67 @@
...
+
+	// Steps into the current value if it is a container. It returns an error if there is no current value or if
+	// the value is not a container. On success, the Reader is positioned before the first value in the container.
+	StepIn() error
+
+	// Steps out of the current container value being read. It returns an error if this Reader is not currently
+	// stepped into a container. On success, the Reader is positioned after the end of the container, but before any
+	// subsequent values in the stream.
+	StepOut() error
+}
```
therapon, line 38: I would expect these APIs to return `(Event, error
)`

PR #32, opened 2020-05-20
Adding support for pretty-printing text Ion

fernomac:

### DIFF
--- a/ion/textwriter.go
+++ b/ion/textwriter.go
@@ -16,13 +16,18 @@ const (
 	// know you're only emiting one datagram; dangerous if there's a chance you're going
 	// to emit another datagram using the same Writer.
 	TextWriterQuietFinish TextWriterOpts = 1
+
+	// TextWriterPretty enables pretty-printing mode.
+	TextWriterPretty TextWriterOpts = 2
 )
 
 // textWriter is a writer that writes human-readable text
 type textWriter struct {
 	writer
-	needsSeparator bool
 	opts           TextWriterOpts
+	needsSeparator bool
+	emptyContainer bool
+	indent         int
 }
 
 // NewTextWriter returns a new text writer.
@@ -255,6 +260,11 @@ func (w *textWriter) Finish() error {
 	return nil
 }
 
+// pretty returns true if we're pretty-printing.
+func (w *textWriter) pretty() bool {
+	return w.opts&TextWriterPretty == TextWriterPretty
+}
+
 // writeValue writes a stringified value to the output stream.
 func (w *textWriter) writeValue(api string, val string) error {
 	if w.err != nil {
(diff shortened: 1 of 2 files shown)


ion/textwriter.go
```diff
@@ -277,47 +288,105 @@ func (w *textWriter) writeValue(api string, val string) error {
...
+
+// writeAnnotations writes out the annotations for a value.
+func (w *textWriter) writeAnnotations() error {
+	as := w.annotations
+	w.annotations = nil
+
+	sep := "::"
+	if w.pretty() {
+		sep = " :: "
+	}
```
therapon, line 382: We do not have a spec for pretty printing. The libs are essentially free to generate something that is easy for humans to read, id valid Ion Text and round trips to the same values. Having said that (and it may be that I am too used to what other Ion libs do) for the annotation's separator typically there are no spaces around `::`. It will make the code also simpler :D

ion/textwriter.go
```diff
@@ -277,47 +288,105 @@ func (w *textWriter) writeValue(api string, val string) error {
...
 		}
-		name := w.fieldName
-		w.fieldName = ""
+	}
 
-		if err := writeSymbol(name, w.out); err != nil {
+	if len(w.annotations) > 0 {
+		if err := w.writeAnnotations(); err != nil {
 			return err
 		}
```
therapon, line 319: `writeAnnotations()` graps the annotations from the writer (which is an array) and ranges over them. I think we can just call `writeAnnotations()` and the case when the annotations are empty will output nothing. Or is my newly found Go Lang knowledge failing me here?

ion/textwriter.go
```diff
@@ -336,9 +407,15 @@ func (w *textWriter) begin(api string, t ctx, c byte) error {
...
+	w.indent++
 	w.needsSeparator = false
+	w.emptyContainer = true
 
-	return writeRawChar(c, w.out)
+	if err := writeRawChar(c, w.out); err != nil {
+		return err
+	}
+
+	return nil
```
therapon, line 418: A quick Google search did not give me an answer here. Can this block of code stay the same, i.e., 

```
return writeRawChar(c, w.out)
```

Is there a reason for checking for the error and returning, as opposed to delegating to the `writeRawChar` function that I believe also returns `nil` or `err`?

ion/textwriter_test.go
```diff
@@ -334,6 +334,69 @@ func TestWriteTextBadFinish(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	actual := buf.String()
+	expected := `{
+	struct : {},
+	list : i :: am :: a :: list :: [
+		"value",
+		null.string
+	],
```
therapon, line 383: List and Structs have `,` separators. Can we add to the test a nesting of these two Types (any combination is fine) and test that pretty print does what we expect for those as well, please.

ion/textwriter.go
fernomac: Hah, this is how I used to "pretty print" Ion by hand back in the day. Happy to follow what the other implementations do, especially if it means I get to delete code. :)

ion/textwriter.go
fernomac: Yeah it would be safe to always call through to `writeAnnotations`. It reads a bit clearer for me like this ("if there are any annotations, write them out"), but I can change it if you prefer.

ion/textwriter.go
fernomac: Ah, yeah, we can just return. I had added some code after the call to `writeRawChar` at some intermediate point, then backed it out but didn't simplify. You'll sometimes see needlessly-explicit error checking like this instead of a straight return so there's less refactoring back and forth as the code changes; I _think_ this part of the code should be relatively stable now though.

ion/textwriter_test.go
fernomac: 👍

PR #33, opened 2020-05-20
Add integration tests with skip lists for failing tests

Armanbqt:
This is integration test with ion-test repository.
This file contains the following tests:
- Roundtrip binary to text and back to binary again
- Roundtrip text to binary and back to text again
- Load bad files (malformed ion files in /bad directory)
- Equivalency test
- Non-equivalency test

Each of the tests have skip list for the tests which are currently failing.
In these tests, low level APIs have been used.


fernomac:
Nice! These are on track to be a lot more comprehensive than the tests in [reader_test.go](https://github.com/amzn/ion-go/blob/master/ion/reader_test.go) and should probably replace those.

ion/integration_test.go
```diff
@@ -0,0 +1,809 @@
...
+
+func skipFile(skipList []string, fn string) bool {
+	ion, _ := regexp.MatchString(`.ion$`, fn)
+	bin, _ := regexp.MatchString(`.10n$`, fn)
+
+	return !ion && !bin || isInSkipList(skipList, fn)
+}
+
+func isInSkipList(skipList []string, fn string) bool {
+	for _, a := range skipList {
```
zslayton, line 554: Since there are hundreds of files and the skip lists are pretty long, scanning the entire skip list for each file could get a bit slow. Can we make `skipList` a `Set` of some kind?

ion/integration_test.go
```diff
@@ -0,0 +1,809 @@
...
+	writeToWriterFromReader(t, r, tw2)
+	tw2.Finish()
+
+	//compare the 2 text writers
+	if !reflect.DeepEqual(tw, tw2) {
+		t.Errorf("Round trip test failed on: " + fp)
+	}
+}
+
+func testLoadBad(t *testing.T, fp string) {
```
zslayton, line 399: Could you add a one-line comment to each of the `test*` helper functions explaining their intent?

ion/integration_test.go
```diff
@@ -0,0 +1,809 @@
...
+	writeToWriterFromReader(t, r, bw2)
+	bw2.Finish()
+
+	// Compare the 2 binary writers
+	if !reflect.DeepEqual(bw, bw2) {
+		t.Errorf("Round trip test failed on: " + fp)
+	}
+}
+
+func textRoundTrip(t *testing.T, fp string) {
```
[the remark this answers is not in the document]
Armanbqt: I originally had separate functions to create the desired writers, but my concern was, that may make the test hard to understand. I Refactored these 2 tests in this commit: https://github.com/amzn/ion-go/pull/33/commits/db465ed3bda1a39bd02f6138f8135d85fba56a4f
Let me know which one you prefer.

ion/integration_test.go
```diff
@@ -0,0 +1,809 @@
...
+	buf := bytes.Buffer{}
+	bw := NewBinaryWriter(&buf)
+	writeToWriterFromReader(t, r, bw)
+	bw.Finish()
+
+	// Make a text writer from the binary writer
+	r = NewReaderBytes(buf.Bytes())
+	str := strings.Builder{}
+	tw := NewTextWriter(&str)
+	writeToWriterFromReader(t, r, tw)
```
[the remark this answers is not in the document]
Armanbqt: It was difficult for me too to name variables only letters!
This code now has been refactored and I addressed the same issue in other functions.

ion/integration_test.go
```diff
@@ -0,0 +1,809 @@
...
+	"reflect"
+	"regexp"
+	"strings"
+	"testing"
+)
+
+// To debug/run one specific file in any of below paths, put the
+// file name with its extension. Works even if the file is listed in skip lists.
+// For example: const debugFile = "ints.ion"
+const debugFile = ""
```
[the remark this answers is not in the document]
Armanbqt: I thought it can be helpful for internal use when we are trying to fix the tests in the skip lists.
Removed.
And I will look into your suggestion for ```--debugFule=```

ion/integration_test.go
```diff
@@ -0,0 +1,809 @@
...
+			w.EndStruct()
+		}
+	}
+
+	if r.Err() != nil {
+		t.Errorf(r.Err().Error())
+	}
+}
+
+func eqv(t *testing.T, r Reader) item {
```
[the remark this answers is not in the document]
Armanbqt: My bad.
Fixed.

ion/integration_test.go
```diff
@@ -0,0 +1,809 @@
...
+	var i item
+
+	an := r.Annotations()
+	if len(an) > 0 {
+		i.annotations = an
+	}
+
+	switch r.Type() {
+	case NullType:
+		i.value = append(i.value, textNulls[NoType])
```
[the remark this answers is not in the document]
Armanbqt: It's from GoLang: [builtin.go#append()](https://golang.org/pkg/builtin/#append)

ion/integration_test.go
```diff
@@ -0,0 +1,809 @@
...
+func loadFile(t *testing.T, path string) []byte {
+	b, err := ioutil.ReadFile(path)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return b
+}
+
+func skipFile(skipList []string, fn string) bool {
+	ion, _ := regexp.MatchString(`.ion$`, fn)
```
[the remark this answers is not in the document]
Armanbqt: Thanks!
```strings.HasSuffix``` is much better and even easier to use.

File: ion/integration_test.go (hunk 1 above)
Comment (Armanbqt) [reply ion/integration_test.go:550]: AFAIK Go does not have ```Set```. The closest thing would be introducing a ```map``` and set filenames as the key. This is only good to make sure we will not have duplicates, but as I was reading there is not a big difference between traversing map versus slice. Seems like slice has a better look up time when the the size is smaller 100 and map can do better for bigger numbers.
Let me know what you think.

File: ion/integration_test.go (hunk 2 above)
Comment (Armanbqt) [reply ion/integration_test.go:395]: Done.

File: ion/integration_test.go (hunk 5 above)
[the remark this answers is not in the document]
Comment (zslayton) [reply ion/integration_test.go]: Cool, thanks.

File: ion/integration_test.go [hunk 9]
```diff
@@ -0,0 +1,807 @@
...
+			r.StepOut()
+		}
+	}
+	if r.Err() != nil {
+		t.Error()
+	}
+}
+
+// Ion values in top-level sequence annotated with "embedded_documents" parse as Ion string and each
+// of the string values in turn, parse as a document and put in an item struct (defined in this file).
```
zslayton, line 463: The `embedded_documents` mechanism is explained pretty succinctly [in the `ion-tests` repo README](https://github.com/amzn/ion-tests). I suggest just copy/pasting it here for clarity:

> If a top-level sequence is annotated with "embedded_documents", it denotes that each of its Ion values is to be parsed as a Ion string value, where its string value is to be parsed as a document. [This function performs that parsing and returns the values in the embedded document as `item`s].

ion/integration_test.go
```diff
@@ -0,0 +1,807 @@
...
+		}
+		newReader := NewReaderStr(str)
+		for newReader.Next() {
+			values = append(values, readCurrentValue(t, newReader))
+		}
+	}
+	return values
+}
+
+func embeddedDoc(an []string) bool {
```
zslayton, line 479: Per the Go style guidelines, I suggest renaming this "isEmbeddedDoc" as it returns a `bool`.

ion/integration_test.go
```diff
@@ -0,0 +1,807 @@
...
+	for i := 0; i < len(values); i++ {
+		for j := i + 1; j < len(values); j++ {
+			if i == j {
+				continue
+			}
+			if eq {
+				if !reflect.DeepEqual(values[i].value, values[j].value) ||
+					!reflect.DeepEqual(values[i].annotations, values[j].annotations) ||
+					!reflect.DeepEqual(values[i].ionType, values[j].ionType) {
+					t.Error("Equivalency test failed. All values should interpret equal.")
```
zslayton, line 499: Could you add the values of `i` and `j` to these error messages so users can tell which values were the source of the issue? If it's easy, including the interpreted values themselves would be helpful too.

Additional phrasing suggestions: "All values should be interpreted as equal." and "Values should not be interpreted as equal."

ion/integration_test.go
```diff
@@ -0,0 +1,807 @@
...
+const badPath = "../ion-tests/iontestdata/bad"
+const equivsPath = "../ion-tests/iontestdata/good/equivs"
+const nonEquivsPath = "../ion-tests/iontestdata/good/non-equivs"
+
+type testingFunc func(t *testing.T, path string)
+
+type item struct {
+	ionType     Type
+	annotations []string
+	value       []interface{}
```
zslayton, line 38: Should `value` be `[]item` to restrict it a bit?

ion/integration_test.go
```diff
@@ -0,0 +1,807 @@
...
+)
+
+const goodPath = "../ion-tests/iontestdata/good"
+const badPath = "../ion-tests/iontestdata/bad"
+const equivsPath = "../ion-tests/iontestdata/good/equivs"
+const nonEquivsPath = "../ion-tests/iontestdata/good/non-equivs"
+
+type testingFunc func(t *testing.T, path string)
+
+type item struct {
```
zslayton, line 35: If we don't already have one, could you open an issue to track replacing this type with something more formal when we get around to it?

ion/integration_test.go
```diff
@@ -0,0 +1,807 @@
...
+	}
+
+	if reader.Err() != nil {
+		t.Errorf(reader.Err().Error())
+	}
+}
+
+// Read the current value in the reader and put that in an item struct (defined in this file).
+func readCurrentValue(t *testing.T, reader Reader) item {
+	var item item
```
zslayton, line 698: Without extra context, this variable declaration was a bit puzzling. Maybe we could rename the internal struct type `ionItem` or `ionValue` for clarity?

ion/integration_test.go
```diff
@@ -0,0 +1,807 @@
...
+
+func equivalencyAssertion(t *testing.T, values []item, eq bool) {
+	// Nested for loops to evaluate each item with all the other items in the list/struct/sexp
+	for i := 0; i < len(values); i++ {
+		for j := i + 1; j < len(values); j++ {
+			if i == j {
+				continue
+			}
+			if eq {
+				if !reflect.DeepEqual(values[i].value, values[j].value) ||
```
zslayton, line 496: It might be helpful to move these `DeepEqual` checks into a method on the `item` type so you can do something like this:

```go
areEqual := values[i].isEqualTo(values[j])
if eq {
  if !areEqual {
    t.Error(...)
} else {
  if areEqual {
    t.Error(...)
  }
}
```

ion/integration_test.go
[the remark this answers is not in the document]
zslayton: Ah, thank you!

ion/integration_test.go
Armanbqt: We cannot. We are adding different types of Ion in an array of this struct in ```readCurrentValue()```, for example ```reader.StringValue()``` or ```reader.TimeValue()``` and they are of type interface{}

ion/integration_test.go
```diff
@@ -0,0 +1,807 @@
...
+	binWriter, buf := makeBinaryWiter(t, fileBytes)
+
+	// Make a text writer from the binary writer
+	_, str := makeTextWriter(t, buf.String())
+
+	// Make another binary writer using writer
+	binWriter2, _ := makeBinaryWiter(t, []byte(str.String()))
+
+	// Compare the 2 binary writers
+	if !reflect.DeepEqual(binWriter, binWriter2) {
```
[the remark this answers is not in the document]
Armanbqt: Are you suggesting to change the comparison between ```buf``` and the buffer for ```binWriter2```?
I tested it and the results are different from current comparison.

ion/integration_test.go
Armanbqt: It does look nicer, but I thought @therapon  suggested to use ```DeepEqual``` for now.

ion/integration_test.go
Armanbqt: Cool, thanks.

ion/integration_test.go
```diff
@@ -0,0 +1,807 @@
...
+	})
+}
+
+func TestNonEquivalency(t *testing.T) {
+	readFilesAndTest(t, nonEquivsPath, nonEquivsSkipList, func(t *testing.T, path string) {
+		testEquivalency(t, path, false)
+	})
+}
+
+// Execute round trip testing for BinaryWriter. Create a BinaryWriter, using the BinaryWriter creat a
```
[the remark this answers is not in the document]
Armanbqt: Fixed. Thanks.

ion/integration_test.go
```diff
@@ -0,0 +1,807 @@
...
+	reader := NewReader(strings.NewReader(data))
+	str := strings.Builder{}
+	txtWriter := NewTextWriter(&str)
+	writeToWriterFromReader(t, reader, txtWriter)
+	txtWriter.Finish()
+	return txtWriter, str
+}
+
+// Create a BinaryWriter from data parameter. Return the writer and buffer containing writer's contents
+func makeBinaryWiter(t *testing.T, data []byte) (Writer, bytes.Buffer) {
```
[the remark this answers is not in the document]
Armanbqt: Renamed.

ion/integration_test.go
Armanbqt: I put the explanations in case someone is looking at test and without knowing ```embedded_documents```.
Modified the comment.

ion/integration_test.go
Armanbqt: Done.

ion/integration_test.go
Armanbqt: [reader struct in Reader.go](https://github.com/amzn/ion-go/blob/master/ion/reader.go#L188) is a supreset of this struct but I did not want to use that.

I do not understand why we need a more formal type for this. ```ionItem``` is being used only for this test, should we create a public type?

ion/integration_test.go
Armanbqt: Renamed to ```ionItem```. Thanks.

ion/integration_test.go
```diff
@@ -0,0 +1,808 @@
...
+	}
+}
+
+// Create a TextWriter from data parameter. Return the writer and string builder containing writer's contents
+func encodeAsTextIon(t *testing.T, data string) (Writer, strings.Builder) {
+	reader := NewReader(strings.NewReader(data))
+	str := strings.Builder{}
+	txtWriter := NewTextWriter(&str)
+	writeToWriterFromReader(t, reader, txtWriter)
+	txtWriter.Finish()
```
matthewh-BQ, line 380: Only caught this through linting, I'm getting a few warnings with unhandled exceptions.

Been looking through best practices for error handling in tests and it seems we can just make this obvious if we're ignoring an error by doing `_ = txtWriter.Finish()`.

Looking into what the best practices are, but I'd feel failing early when an error occurred would be good. Let says if the text writer failed to finish, would this be obvious if we continued to ignoring the error and let the test fail later on? But, this could be a unit test vs integration test scenario, so maybe explicity ignoring the error for these cases is good enough.

Thoughts?

Armanbqt: Correct, I did not put error handling for ```writer.Finish()```, ```stepIn()``` and ```stepOut()```.
My approach was to test the testfiles against different scenarios of this file (roundtrip, equivalency, loadBad). And if it fails at any point, the test fails. But open to suggestions.

ion/integration_test.go
zslayton: Ah, ok. Thanks for explaining.

ion/integration_test.go
fernomac: My general preference would be to explicitly check for errors from the writer and `t.Fatal(err)`; if something goes wrong during writing, that'll give you more information about the problem than one of the outputs being truncated.

The writers _should_ remember any errors they encounter along the way, no-op future writes, and return the error from `Finish`, so in theory you can get away with just checking the return value there. (Although that's maybe something that's worth testing more explicitly than it is tested now. :))

ion/integration_test.go
zslayton: > I do not understand why we need a more formal type for this. ionItem is being used only for this test, should we create a public type?

Sorry, to clarify: I don't think we need a more formal type for this--it's fine as-is. However, I expect that at some point in the (distant?) future, `ion-go` will have a public-facing formal type that represents any Ion value. (Similar to `ion-java`'s `IonValue` class.) If/when we add that, we should replace this so we're not maintaining two data types for the same thing.

ion/integration_test.go
zslayton: Sure, that's fine.

ion/integration_test.go
```diff
@@ -496,13 +495,15 @@ func equivalencyAssertion(t *testing.T, values []item, eq bool) {
 				if !reflect.DeepEqual(values[i].value, values[j].value) ||
 					!reflect.DeepEqual(values[i].annotations, values[j].annotations) ||
 					!reflect.DeepEqual(values[i].ionType, values[j].ionType) {
-					t.Error("Equivalency test failed. All values should interpret equal.")
+					t.Errorf("Equivalency test failed. All values should be interpreted as"+
```
zslayton, line 498: Looks like you need a space after "as" here and on line 505.

ion/integration_test.go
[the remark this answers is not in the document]
zslayton: Per our offline discussion, we want to compare the values that can be read from each encoding rather than comparing the `Writer`s or buffers directly.

ion/integration_test.go
Armanbqt: Created an issue for that: #45

ion/integration_test.go
```diff
@@ -0,0 +1,808 @@
...
+			if err != nil {
+				t.Errorf("Something went wrong when reading Boolean value. " + err.Error())
+			}
+			err = writer.WriteBool(val)
+			if err != nil {
+				t.Errorf("Something went wrong when writing Boolean value. " + err.Error())
+			}
+
+		case IntType:
+			val, err := reader.Int64Value()
```
matthewh-BQ, line 590: Caught this while working on int support, I believe this should be BigIntValue() and then we also write bigint.

ion/integration_test.go
```diff
@@ -0,0 +1,808 @@
...
+	case BoolType:
+		val, err := reader.BoolValue()
+		if err != nil {
+			t.Errorf("Something went wrong when reading Boolean value. " + err.Error())
+		}
+		ionItem.value = append(ionItem.value, val)
+		ionItem.ionType = BoolType
+
+	case IntType:
+		val, err := reader.Int64Value()
```
matthewh-BQ, line 720: Should this be BigIntValue()?

ion/integration_test.go
Armanbqt: Oh I see. Thanks for the explanation. We also had this conversation internally. Glad you brought that up too.
Issue #46

ion/integration_test.go
```diff
@@ -40,47 +40,51 @@ type ionItem struct {
...
 	"testfile22.ion",
-	"testfile23.ion",
-	"testfile31.ion",
-	"testfile35.ion",
-	"testfile37.ion",
-	"utf16.ion",
-	"utf32.ion",
-	"whitespace.ion",
+	"T5.10n", /*
+		"bigInts.ion",
```
zslayton, line 47: Do these tests no longer fail using the new `Reader`-based roundtrip test?

ion/integration_test.go
```diff
@@ -357,38 +368,45 @@ func testTextRoundTrip(t *testing.T, fp string) {
...
 
-	// Re-encode buf's stream as text in another text writer
-	txtWriter2, _ := encodeAsTextIon(t, buf.String())
+	reader1 := NewReader(strings.NewReader(str.String()))
+	reader2 := NewReader(bytes.NewReader(buf.Bytes()))
 
-	//compare the 2 text writers
-	if !reflect.DeepEqual(txtWriter, txtWriter2) {
-		t.Errorf("Round trip test failed on: " + fp)
+	for reader1.Next() {
```
zslayton: This looks good. 👍

ion/integration_test.go
```diff
@@ -346,19 +346,10 @@ func testBinaryRoundTrip(t *testing.T, fp string) {
...
-		i2 := readCurrentValue(t, reader2)
-
-		if !reflect.DeepEqual(i1.value, i2.value) ||
-			!reflect.DeepEqual(i1.annotations, i2.annotations) ||
-			!reflect.DeepEqual(i1.ionType, i2.ionType) {
-			t.Errorf("failed %s %v vs %v", fp, i1.value, i2.value)
-		}
+	buf2 := encodeAsBinaryIon(t, []byte(str.String()))
+
+	if !reflect.DeepEqual(buf, buf2) {
```
zslayton, line 351: We don't want to compare the buffers or strings that are produced directly; the encoded bytes can be different but still contain equivalent values. For example, the `TextWriter` and the `PrettyTextWriter` will produce different `string`s, but the values in the strings should be considered the same. Similarly, some data types like `Timestamp` have a variety of ways to encode the same value; `2007-02-23T00:00Z` and `2007-02-23T00:00+00:00` are equivalent despite being encoded differently.

ion/integration_test.go
Armanbqt: Changed.

ion/integration_test.go
Armanbqt: Changed. Thanks.

ion/integration_test.go
Armanbqt: Updated the list

ion/integration_test.go
Armanbqt: I was hoping you read the email I sent about this before seeing it in the PR :P
Updated the test as talked yesterday.

ion/integration_test.go
Armanbqt: Done.

ion/integration_test.go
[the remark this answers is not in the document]
Armanbqt: Done

ion/integration_test.go
```diff
@@ -0,0 +1,923 @@
...
+		i2 := readCurrentValue(t, reader2)
+
+		if !i1.equal(i2) {
+			t.Errorf("Failed on %s round trip. Binary reader has %v "+
+				"where the value in Text reader is %v", fp, i1.value, i2.value)
+		}
+	}
+}
+
+// Execute round trip testing for TextWriter. Create a TextWriter, using the TextWriter creat a
```
zslayton, line 372: * It looks like these comments are out of date now. ("Validate that the first and last Writers are equal.")
* "creat" -> "create"

ion/integration_test.go
```diff
@@ -0,0 +1,923 @@
...
+		i2 := readCurrentValue(t, reader2)
+
+		if !i1.equal(i2) {
+			t.Errorf("Failed on %s round trip. Text reader has %v "+
+				"where the value in Binary reader is %v", fp, i1.value, i2.value)
+		}
+	}
+}
+
+// Create a TextWriter from data parameter. Return the writer and string builder containing writer's contents
```
zslayton, line 398: These function comments refer to the old return types.

ion/integration_test.go
```diff
@@ -0,0 +1,923 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	return str
+}
+
+// Create a BinaryWriter from data parameter. Return the writer and buffer containing writer's contents
+func encodeAsBinaryIon(t *testing.T, data []byte) bytes.Buffer {
+	reader := NewReader(bytes.NewReader(data))
+	buf2 := bytes.Buffer{}
```
zslayton, line 414: It looks like `buf2` could be named `buf`?

ion/integration_test.go
```diff
@@ -343,8 +343,8 @@ func TestNonEquivalency(t *testing.T) {
 	})
 }
 
-// Execute round trip testing for BinaryWriter. Create a BinaryWriter, using the BinaryWriter create
-// a TextWriter and back to BinaryWriter again. Validate that the first and last Writers are equal.
+// Execute round trip testing for the BinaryWriter by creating a BinaryWriter and re-encoding it to text.
+// From the output of the writers, construct two readers and verify the equivalency of both readers.
```
zslayton, line 347:
```
// Re-encodes the provided file as binary Ion, reads that binary Ion and writes it as text Ion,
// then constructs Readers over the binary and text encodings to verify that the streams 
// are equivalent.
```

ion/integration_test.go
```diff
@@ -369,8 +369,8 @@ func testBinaryRoundTrip(t *testing.T, fp string) {
 	}
 }
 
-// Execute round trip testing for TextWriter. Create a TextWriter, using the TextWriter creat a
-// BinaryWriter and back to TextWriter again. Validate that the first and last Writers are equal.
+// Execute round trip testing for the TextWriter by creating a TextWriter and re-encoding it to binary.
+// From the output of the writers, construct two readers and verify the equivalency of both readers.
```
zslayton, line 373:
```
// Re-encodes the provided file as text Ion, reads that text Ion and writes it as binary Ion, 
// then constructs Readers over the text and binary encodings to verify that the streams 
// are equivalent.
```

ion/integration_test.go
```diff
@@ -395,7 +395,7 @@ func testTextRoundTrip(t *testing.T, fp string) {
 	}
 }
 
-// Create a TextWriter from data parameter. Return the writer and string builder containing writer's contents
+// Create a TextWriter from data parameter and return the string builder containing writer's contents.
```
zslayton, line 398: Now that we're not returning the `TextWriter`, it's an implementation detail.
```
// Re-encode the provided Ion data as a text Ion string.
```

ion/integration_test.go
```diff
@@ -408,17 +408,17 @@ func encodeAsTextIon(t *testing.T, data string) strings.Builder {
 	return str
 }
 
-// Create a BinaryWriter from data parameter. Return the writer and buffer containing writer's contents
+// Create a BinaryWriter from data parameter and return the buffer containing writer's contents.
```
zslayton, line 411: Now that we're not returning the `BinaryWriter`, it's an implementation detail.
```
// Re-encode the provided Ion data as a binary Ion buffer.
```

ion/integration_test.go
```diff
@@ -395,7 +395,7 @@ func testTextRoundTrip(t *testing.T, fp string) {
 	}
 }
 
-// Create a TextWriter from data parameter. Return the writer and string builder containing writer's contents
+// Create a TextWriter from data parameter and return the string builder containing writer's contents.
 func encodeAsTextIon(t *testing.T, data string) strings.Builder {
```
zslayton, line 399: Golang allows you to use the `string` data type to mean "arbitrary bytes", but I don't think we should use it that way. When I first read this function I was surprised because I thought if `data` was a `string`, it must already be text Ion so there'd be no need to encode it as text Ion. Only when I went and looked at the callsite did I realize that we were taking a binary buffer and calling `String()` on it to get a `string` instance.

We should make sure we're only using `string` to refer to text throughout the APIs. If this is a quick change, you can include it in the PR. Otherwise, please open an issue so we don't forget to address it.

Armanbqt: Fixed it in this test, create #48 to check throughout the project.

Issue #34, opened 2020-05-21
Implement isScalar() and isContainer() in type.go


justing-bq:
Can we get the following functions defined in type.go:

isScalar() should return true if the type is NullType, BoolType, IntType, FloatType, DecimalType, TimestampType, SymbolType, StringType, ClobType, or BlobType. Returns false otherwise.

isContainer() should return true if the type is ListType, SexpType, or StructType. Returns false otherwise.

Armanbqt:
Similar logic in [ion-dotnet](https://github.com/amzn/ion-dotnet/blob/master/Amazon.IonDotnet/IonType.cs#L67)

PR #37, opened 2020-05-22
Implement isScalar() and isContainer() in type.go

justing-bq:
Resolves https://github.com/amzn/ion-go/issues/34

### DIFF
--- a/ion/type.go
+++ b/ion/type.go
@@ -92,22 +92,12 @@ func (t Type) String() string {
 
 // IsScalar determines if the type is a scalar type
 func (t Type) IsScalar() bool {
-	switch t {
-	case NullType, BoolType, IntType, FloatType, DecimalType, TimestampType, SymbolType, StringType, ClobType, BlobType:
-		return true
-	}
-
-	return false
+	return NullType <= t && t <= BlobType
 }
 
 // IsContainer determines if the type is a container type
 func (t Type) IsContainer() bool {
-	switch t {
-	case ListType, SexpType, StructType:
-		return true
-	}
-
-	return false
+	return ListType <= t && t <= StructType
 }
 
 // IntSize represents the size of an integer.
--- a/ion/type.go
+++ b/ion/type.go
@@ -91,12 +91,12 @@ func (t Type) String() string {
 }
 
 // IsScalar determines if the type is a scalar type
-func (t Type) IsScalar() bool {
+func IsScalar(t Type) bool {
 	return NullType <= t && t <= BlobType
 }
 
 // IsContainer determines if the type is a container type
-func (t Type) IsContainer() bool {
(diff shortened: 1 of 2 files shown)


ion/type.go
```diff
@@ -90,6 +90,26 @@ func (t Type) String() string {
 	}
 }
 
+// IsScalar determines if the type is a scalar type
+func (t Type) IsScalar() bool {
+	switch t {
+	case NullType, BoolType, IntType, FloatType, DecimalType, TimestampType, SymbolType, StringType, ClobType, BlobType:
```
byronlin-bq, line 96: Instead of the switch statement, we can just check the type against the int value of the enum

Armanbqt:
FYI you can change the description from #34 to _resovles #34_ , so once this gets merged, issue 34 will automatically closed.

ion/type.go
justing-bq: Done.

ion/type.go
```diff
@@ -90,6 +90,16 @@ func (t Type) String() string {
 	}
 }
 
+// IsScalar determines if the type is a scalar type
+func (t Type) IsScalar() bool {
```
fernomac, line 94: Go tends to prefer top-level functions for these; e.g. `strings.HasPrefix(s, "...")` rather than `s.HasPrefix("...")`. That'd be my recommendation (which you're free to ignore if you prefer it this way :)).

ion/type_test.go
```diff
@@ -19,3 +19,32 @@ func TestIntSizeToString(t *testing.T) {
 		}
 	}
 }
+
+func TestIsScalar(t *testing.T) {
+	if NoType.IsScalar() {
+		t.Errorf("Expected IsScalar() to return false for type NoType")
+	}
+	for i := NullType; i <= BlobType; i++ {
```
fernomac, line 27: Might be safer to define the expected behavior here more explicitly, as a guard against someone reordering the Type enum values?

ion/type.go
justing-bq: Done.

ion/type_test.go
justing-bq: Done.

ion/type_test.go
```diff
@@ -19,3 +19,93 @@ func TestIntSizeToString(t *testing.T) {
 		}
 	}
 }
+
+func TestIsScalar(t *testing.T) {
+	if IsScalar(NoType) {
```
zslayton, line 24: To eliminate some of the copy/pasting, I suggest making an array of the scalar types and looping over them. The same could be done for `TestIsContainer`.

I'm new to Go, so if there are reasons that's ill advised, please let me know.

justing-bq: Done.

Issue #44, opened 2020-05-22
Detect malformed Timestamp


Armanbqt:
Malformed timestamp values should throw an exception and not form a valid Reader.

Armanbqt:
Closing, as #50 is merged and fixes malformed Ion timestamps

PR #50, opened 2020-06-03
Invalid Ion Timestamps

Armanbqt:
Issue #44 
Fix malformed Ion timestamps.

Fixes include:
- Year, month, day must be larger than 0 (minimum date is 0001-01-01)
- Months cannot be larger than 12.
- Days, depending on which month, should be either 28,29,30 or 31.
- All different valid formats should be parsed correctly.
- Timestamps with Hour, must have minute too.
- Parse leap years properly.

This fix only aims invalid timestamps. There are still issues with timezone and fractional seconds which are not part of this fix (marked in the code as TODO)

The only outstanding invalid timestamp test in `malformedIonsSkipList` is `timestampLenTooLarge.10n`


ion/bitstream.go
```diff
@@ -522,8 +529,18 @@ func (b *bitstream) ReadTimestamp() (time.Time, error) {
...
 	b.clear()
 
-	utc := time.Date(ts[0], time.Month(ts[1]), ts[2], ts[3], ts[4], ts[5], int(nsecs), time.UTC)
-	return utc.In(time.FixedZone("fixed", int(offset)*60)), nil
+	// TODO: Timezone should not be hard coded to time.UTC
+	return tryCreateTimeWithNSecAndOffset(ts, nsecs, offset, time.UTC)
+}
+
+func tryCreateTimeWithNSecAndOffset(ts []int, nsecs int, offset int64, loc *time.Location) (time.Time, error) {
+	date := time.Date(ts[0], time.Month(ts[1]), ts[2], ts[3], ts[4], ts[5], nsecs, loc)
```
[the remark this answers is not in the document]
Armanbqt: Thanks, David. I think I have to read a bit more about Ion binary timestamp.

ion/textutils.go
```diff
@@ -342,39 +342,97 @@ func parseTimestamp(val string) (time.Time, error) {
...
 		return invalidTimestamp(val)
 	}
-	if val[16] != ':' {
-		return time.Parse("2006-01-02T15:04Z07:00", val)
+	if val[16] == 'z' || val[16] == 'Z' {
+		return time.Parse("2006-01-02T15:04Z", val)
+	}
+	if val[16] == '+' || val[16] == '-' {
+		if isValidOffset(val, 16) {
+			// TODO: we should use time.ParseInLocation()
```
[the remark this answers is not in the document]
Armanbqt: I was thinking to implement a logic where:
offset == 0  ```loc, _ := time.LoadLocation("UTC")```
offset  >  0   ```loc, _ := time.LoadLocation("Local")```
offset == -00:00  ```non - no location```

ion/bitstream.go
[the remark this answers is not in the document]
zslayton: Here's the relevant bit of [the binary `decimal` spec](http://amzn.github.io/ion-docs/docs/binary.html#6-timestamp):
> Timestamp representations have 7 components, where 5 of these components are optional depending on the precision of the timestamp ..... All of these 7 components are in Universal Coordinated Time (UTC).
> .....
> Note: The component values in the binary encoding are always in UTC, while components in the text encoding are in the local time! This means that transcoding requires a conversion between UTC and local time.

ion/textutils.go
```diff
@@ -342,39 +342,97 @@ func parseTimestamp(val string) (time.Time, error) {
...
 	}
 
+	// At this point timestamp must have hour:minute
 	if len(val) < 17 {
 		return invalidTimestamp(val)
 	}
-	if val[16] != ':' {
-		return time.Parse("2006-01-02T15:04Z07:00", val)
+	if val[16] == 'z' || val[16] == 'Z' {
+		return time.Parse("2006-01-02T15:04Z", val)
```
zslayton, line 362: As a Go newbie, this line really confused me. I had to read [the `Parse()` docs](https://golang.org/pkg/time/#Parse) to understand why we were writing out a seemingly arbitrary date from the winter of 2006.

> Parse parses a formatted string and returns the time value it represents. The layout defines the format by showing how the reference time, defined to be
> 
>     Mon Jan 2 15:04:05 -0700 MST 2006
> 
> would be interpreted if it were the value; it serves as an example of the input format. The same interpretation will then be made to the input string. 
>
>  Predefined layouts ANSIC, UnixDate, **RFC3339** and others describe standard and convenient representations of the reference time. For more information about the formats and the definition of the reference time, see the documentation for ANSIC and the other constants defined by this package. Also, the executable example for Time.Format demonstrates the working of the layout string in detail and is a good reference. 

I'd like to use `time.RFC3339` instead of the string literal `"2006-01-02T15:04Z"` for simplicity.

ion/textutils.go
```diff
@@ -342,39 +342,97 @@ func parseTimestamp(val string) (time.Time, error) {
...
-		if i >= 29 {
-			// Too much precision for a go Time.
-			// TODO: We should probably round instead of truncating? Ah well.
-			return time.Parse(time.RFC3339Nano, val[:29]+val[i:])
+		if val[idx] == '+' || val[idx] == '-' {
+			if isValidOffset(val, idx) {
+				if idx >= 29 {
+					// Too much precision for a go Time.
+					// TODO: We should probably round instead of truncating? Ah well.
+					// TODO: we should use time.ParseInLocation()
```
zslayton, line 399: Same thoughts here RE: `ParseInLocation`.

ion/textutils.go
```diff
@@ -342,39 +342,97 @@ func parseTimestamp(val string) (time.Time, error) {
...
+			}
+			return invalidTimestamp(val)
 		}
 	}
+	return invalidTimestamp(val)
+}
+
+func tryCreateTimeDate(val string, year int64, month int64, day int64, loc *time.Location) (time.Time, error) {
+	date := time.Date(int(year), time.Month(month), int(day), 0, 0, 0, 0, loc)
+	// time.Date converts 2000-01-32 input to 2000-02-01
```
zslayton, line 413: Does `time.Date` convert negative numbers to the previous month? The docs are unclear.

> The month, day, hour, min, sec, and nsec values may be outside their usual ranges and will be normalized during the conversion. For example, October 32 converts to November 1.

If it does, this check could also handle all of the `if unit < 1` checks that appear above.

ion/textutils.go
```diff
@@ -342,39 +342,97 @@ func parseTimestamp(val string) (time.Time, error) {
...
+		return false
+	}
+
+	hourOffset, errH := strconv.ParseInt(val[idx+1:idx+3], 10, 32)
+	minuteOffset, errM := strconv.ParseInt(val[idx+4:], 10, 32)
+	if errH != nil || errM != nil {
+		return false
+	}
+
+	// TODO: check the offset against max offset (14 * 60), adjust and return if larger than max
```
zslayton, line 432: Per our earlier conversation, I don't think we should change the provided offset unless RFC3339 (on which `timestamp` is based) says that there's a max.

ion/textutils.go
zslayton: If it turns out that `time.RFC3339` behaves differently from the string literal version, a comment explaining what the string literal means would also be fine.

fernomac: Yeah, it's a clever way of doing format strings and also _totally_ confusing if you're not aware of that convention. :/ `time.RFC3339` expects the string being parsed to include seconds, which is why this one was a literal. I'd be +1 for adding a comment explaining what it's doing.

ion/textutils.go
Armanbqt: Good catch, I missed it. However, we needed ```year < 1``` to protect against ```0000T``` or ```0000-02-02T```.

ion/textutils.go
Armanbqt: Added some explanation.
__  
We also can add something like:
```
const (
	Layout_yyyy_mm_ddThh_mmZhh_mm         = "2006-01-02T15:04Z07:00"   // Comment
	Layout_yyyy_mm_ddThh_mmZ              = "2006-01-02T15:04Z"             // Comment
	Layout_yyyy_mm_ddThh_mm_ss_nsecZhh_mm = time.RFC3339Nano // "2006-01-02T15:04:05.999999999Z07:00"
	Layout_yyyy_mm_ddThh_mm_ssZhh_mm      = time.RFC3339     // "2006-01-02T15:04:05Z07:00"
)
```

and then use these constants: ``` return time.Parse(Layout_yyyy_mm_ddThh_mmZhh_mm, val) ```

Let me know which one is preferred.

ion/bitstream.go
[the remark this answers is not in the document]
Armanbqt: Thank you guys

Armanbqt:
Thanks Zack and David for the review.
Updated the pull request.

ion/textutils.go
zslayton: Creating named constants for this is a good idea. :thumbsup:

ion/textutils.go
```diff
@@ -9,6 +9,24 @@ import (
 	"time"
 )
 
+const (
+	// LayoutYYYYMMDDTHHMMZHHMM time.date format. time.Parse() uses
+	// "2006-01-02T15:04Z07:00" explicitly as a string value to identify this format.
+	LayoutYYYYMMDDTHHMMZHHMM = "2006-01-02T15:04Z07:00"
```
Armanbqt, line 15: We cannot use ```_``` in constant names (linter).
I am open to suggestions if anyone can think of better names.

zslayton: > We cannot use _ in constant names (linter).

Welp, it only took us a few hours to stub our collective toe on the linter. 😛 

At a glance, it's pretty hard to tell the difference between the names. `LayoutYYYYMMDDTHHMMZHHMM` and `LayoutYYYYMMDDTHHMMZ` are visually very similar. Maybe there's a good name set that emphasizes what's different about each option? (i.e. the resolution/suffixes)

How about:
```go
LayoutMinutesZulu // ('Z')
LayoutMinutesAndOffset // ('ZHHMM')
LayoutSecondsAndOffset  // RFC3339
LayoutNanosecondsAndOffset // RFC3339Nano
```
?

fernomac: You presumably want them lower-case so they're not part of the public API?

My vote would be to use the `time.` constants directly, and add a local definition for `const rfc3339Minute = "2006-01-02T15:04Z07:00"`.

Armanbqt: Thanks David! That was an oversight on my part - still sometimes I forget Capital letter means Public!
However, we might keep ```LayoutMinutesAndOffset``` naming as it is more descriptive than ```RFC3339``` for a person who is as clueless as me on what RFC3339 means at first glance.

Issue #36, opened 2020-05-22
Fix Decimal type issues


Armanbqt:
For values around decimal bounds (like ones in [decimal64BitBoundary](https://github.com/amzn/ion-tests/blob/05ecd5de78ac6244826cd4e77bae932925c966c1/iontestdata/good/decimal64BitBoundary.ion)), or big values of decimal, and decimals with underscore we cannot have a successful round trip.
Doing this should result in removing tests regarding Decimal from the skip lists.

matthewh-BQ:
@zslayton : I believe this bucket task can be closed.

Looked more into the failing tests with the PR #51 , remaining of failing tests associated with issue #67.

Decimal support considers precision and different representations. For example:

Not Equal
```
// different precisions (negative zeroes)
[
    -0.,
    -0.0,
    -0.00,
    -0.000000,
]
```

Equal
```
(
    0.
    0d0
)
```

Armanbqt:
Closing in favor of #67

PR #51, opened 2020-06-04
FIX Decimal 'D' format scale 0 being omitted in binary

matthewh-BQ:
Issue 36:
Description of Changes:

Decimal values being represented in the 'D' format that have a scale of 0
were being omitted in the binary slice representation. This matches the
behaviour in the other Ion libraries. The issue was being caused because
with the omission of the scale 0, the dot binary representation in the slice was
being mistaken as the scale value.


matthewh-BQ:
@fernomac : If possible, could we please get some historical context to the intention with the conditional check for ensuring that exp != 0?

We've seen some issues with the round trip testing specifically with decimals following the 'D' format with a scale of 0. e.g., 123456d0

When the decimal is being written, the buffer does not factor it in.

The unit test was changed to match what we think ought to be happening.

fernomac:
Ah, I think this was a misread of the spec on my part:

“ Decimal representations have two components: exponent (a VarInt) and coefficient (an Int). The decimal’s value is coefficient * 10 ^ exponent.

The length of the coefficient subfield is the total length of the representation minus the length of exponent. The subfield should not be present (that is, it has zero length) when the coefficient’s value is (positive) zero.

If the value is 0. (aka 0d0) then L is zero, there are no length or representation fields, and the entire value is encoded as the single byte 0x50.”

So a zero exponent should be omitted if and only if the coefficient is also zero?

ion/binarywriter_test.go
```diff
@@ -175,7 +175,7 @@ func TestWriteBinaryTimestamp(t *testing.T) {
 
 func TestWriteBinaryDecimal(t *testing.T) {
 	eval := []byte{
-		0x50,       // 0.
```
fernomac, line 178: Might want to leave this in for 0. since that seems to be specified in the spec?

matthewh-BQ: Yes, that's right. Missed that in the spec and you explained it well! Thank you!

matthewh-BQ:
> Ah, I think this was a misread of the spec on my part:
> 
> “ Decimal representations have two components: exponent (a VarInt) and coefficient (an Int). The decimal’s value is coefficient * 10 ^ exponent.
> 
> The length of the coefficient subfield is the total length of the representation minus the length of exponent. The subfield should not be present (that is, it has zero length) when the coefficient’s value is (positive) zero.
> 
> If the value is 0. (aka 0d0) then L is zero, there are no length or representation fields, and the entire value is encoded as the single byte 0x50.”
> 
> So a zero exponent should be omitted if and only if the coefficient is also zero?

Yes, that makes sense. I've added a conditional for this special case.

ion/binarywriter.go
```diff
@@ -172,21 +172,24 @@ func (w *binaryWriter) WriteFloat(val float64) error {
...
-	}
-	if coef.Sign() != 0 {
-		vlen += bigIntLen(coef)
+	// If the value is 0. (aka 0d0) then L is zero, there are no length or
+	// representation fields, and the entire value is encoded as the single byte 0x50.
+	if coef.Sign() == 0 && int64(exp) == 0 {
+		buf := make([]byte, 0, 0)
+		buf = appendTag(buf, 0x50, 0)
+
+		return w.writeValue("Writer.WriteDecimal", buf)
```
fernomac, line 181: You can just use `w.writeTag(0x50, 0)` here. :) The more complicated path of allocating a slice and using appendXxx to fill it in potentially saves some memory allocations if you can calculate ahead of time how big the buf needs to be.

(Which may or may not actually be any faster, someone should probably profile that...)

zslayton:
Related to issue #36.

PR #54, opened 2020-06-05
use io.ReadFull when reading binary values of known length

### LINKED ISSUE #53: Issue when attempting to deserialize multiple integer values from a binary ion file
Hi, I'm seeing the following issue when attempting to deserialize multiple 8-byte integer values from a binary ion file:

Example steps to reproduce:

1. Serialize the same value 10,000 times using the following python script:
```python
from amazon.ion import simpleion

file = open("dummy.ion", "wb+")

for i in range(10000):
    simpleion.dump(1570737066801085, file, binary = True)
```

2. Attempt to deserialize the integers using ion-go:
```go
package main

import (
"github.com/amzn/ion-go/ion"
"os"
)
func main() {
        f, err := os.Open("dummy.ion")
        if err != nil { panic(err) }
        defer f.Close()
        dec := ion.NewDecoder(ion.NewReader(f))
        enc := ion.NewTextEncoder(os.Stdout)

        for {
                val, err := dec.Decode()
                if err == ion.ErrNoInput { break } else if err != nil { panic(err) }
                if err := enc.Encode(val); err != nil { panic(err) }
        }

        if err := enc.Finish(); err != nil { panic(err) }
}
```

This will result in the first 510 values being read successfully, but for the 511st value I see the following error:
```
1570737066801085
1570737066801085
1570737066801085
1570737066801085
1570733964656640panic: ion: syntax error: malformed annotation (offset 8194)
```

Am I doing something wrong with the use of ion-go in the code snippet above?

Notes:
- I used `go get -u github.com/amzn/ion-go/ion` to import ion-go

- Deserializing the values with ion-python is successful:
```python
from amazon.ion import simpleion

file = open("dummy.ion", "rb")
data = simpleion.load(file, single_value = False)

for d in data:
    print(d)
```
- Serializing using ion-c will also result in the same error when reading with ion-go, but it will be successful when deserialized using ion-python

- Serializing the integers from within a list and then deserializing it with the ion-go code above will result in the list being deserialized successfully
```python
file = open("dummy.ion", "wb+")
data = [1570737066801085] * 10000
simpleion.dump(data, file, binary = True)
```

fernomac:
Resolves #53 

Currently, when reading a binary value whose length we already know, such as an integer, we call `in.Read` with a slice of the appropriate length. The contract of `io.Reader` allows `Read` to read fewer bytes than requested even if we have not reached the end of the stream--for example, when reading from a `bufio.Reader` whose internal buffer has _some_ bytes, but not enough to satisfy the entire read...

This fixes that issue by switching to use `io.ReadFull`, which will loop around and read some more until the entire buffer has been filled or EOF has been reached. I did a quick grep and this is the only place in the library where we're using `Reader.Read`.

### DIFF
--- a/ion/unmarshal_test.go
+++ b/ion/unmarshal_test.go
@@ -537,3 +537,28 @@ func TestDecode(t *testing.T) {
 	test("()", []interface{}{})
 	test("(1 + two)", []interface{}{1, "+", "two"})
 }
+
+func TestDecodeLotsOfInts(t *testing.T) {
+	// Regression test for https://github.com/amzn/ion-go/issues/53
+	buf := bytes.Buffer{}
+	w := NewBinaryWriter(&buf)
+	for i := 0; i < 512; i++ {
+		w.WriteInt(1570737066801085)
+	}
+	w.Finish()
+	bs := buf.Bytes()
+
+	dec := NewDecoder(NewReaderBytes(bs))
+	for {
+		val, err := dec.Decode()
+		if err == ErrNoInput {
+			break
+		}
+		if err != nil {
+			t.Fatal(err)
+		}
+		if val.(int64) != 1570737066801085 {
+			t.Fatalf("expected %v, got %v", 1570737066801085, val)
+		}
+	}
+}
--- a/ion/bitstream.go
+++ b/ion/bitstream.go
@@ -837,10 +837,10 @@ func (b *bitstream) readN(n uint64) ([]byte, error) {
 	}
 
 	bs := make([]byte, n)
-	actual, err := b.in.Read(bs)
+	actual, err := io.ReadFull(b.in, bs)
 	b.pos += uint64(actual)
 
-	if err == io.EOF {
+	if err == io.EOF || err == io.ErrUnexpectedEOF {
 		return nil, &UnexpectedEOFError{b.pos}
 	}
 	if err != nil {


ion/unmarshal_test.go
```diff
@@ -537,3 +537,28 @@ func TestDecode(t *testing.T) {
 	test("()", []interface{}{})
 	test("(1 + two)", []interface{}{1, "+", "two"})
 }
+
+func TestDecodeLotsOfInts(t *testing.T) {
```
zslayton, line 541: Could you help me understand how this test exercises the case where `Read()` would return fewer bytes than were requested?

fernomac: Under the covers we create a bufio.Reader with a default (4096-byte) internal buffer. After reading 4 bytes of BVM and 511 x 8 byte integer values, there are 4 bytes remaining in that buffer, plus another 4 in the underlying stream. Pre-fix, we’d do a partial read of those four bytes as a (corrupt) 512th integer, then try to read the remaining four bytes as another value.

(Might be worth trying to add a more specific test with a custom Reader so it’s more obvious what’s going on, if the bufio.Reader doesn’t get in the way? This is the code I wrote to debug the issue and figured I might as well include :))

zslayton: Ah, perfect! Thank you. Would you mind adding a comment to the test explaining the buffer size math?

> Might be worth trying to add a more specific test with a custom Reader so it’s more obvious what’s going on, if the bufio.Reader doesn’t get in the way?

Agreed. I'll open an issue to track this.

PR #58, opened 2020-06-06
Fix for string values

Armanbqt:
This pull request contains fixes for 2 cases:
- Single quote in a long string: ```''' ion's fun! '''```
- Strings with control characters


ion/tokenizer.go
```diff
@@ -668,6 +678,27 @@ func (t *tokenizer) readEscapedChar(clob bool) (rune, error) {
 	return 0, &SyntaxError{fmt.Sprintf("bad escape sequence '\\%c'", c), t.pos - 2}
 }
 
+func (t *tokenizer) isProhibitedControlChar(c int) bool {
```
fernomac, line 681: t is unused here, so maybe worth dropping it and adding these functions to textutils.go?

Armanbqt:
Thanks @fernomac , I applied your suggestions and updated the PR.

I do not have a preference between merge versus  rebase (even though as you mentioned, rebase keeps the history cleaner).
Which ever we decide, I am fine with it. Any thoughts on this @zslayton ?

zslayton:
> I do not have a preference between merge versus rebase (even though as you mentioned, rebase keeps the history cleaner).
> Which ever we decide, I am fine with it. Any thoughts on this @zslayton ?

My preference is to rebase while the PR is a draft and merge once the first review has happened. This means that reviewers will see a version of the code that was up-to-date with `master` when the (non-draft) PR was opened, but that the discussion around the code doesn't get blown away by rebasing after comments have been added.

In short, if you have merges in your commit history when you go to request your first review, rebase to clean it up. When the PR is approved, we'll merge from master and re-review if large changes were required. Finally, we'll squash+merge back into `master`.

ion/tokenizer.go
```diff
@@ -582,20 +585,27 @@ func (t *tokenizer) readLongString() (string, error) {
...
+			startPosition := t.pos
 			ok, err := t.skipEndOfLongString(t.skipCommentsHandler)
 			if err != nil {
 				return "", err
 			}
 			if ok {
 				return ret.String(), nil
 			}
-
+			if startPosition == t.pos {
```
zslayton, line 605: It looks like [`skipEndOfLongString` reports whether it consumed the string ending](https://github.com/amzn/ion-go/blob/64990f801f0f53d08a2d7cdb6f474c5a19fbb5ff/ion/skipper.go#L534-L537), so I think you can replace this position check with an `else {...}` attached to the `if ok {...}`.

```go
wasConsumed, err := t.skipEndOfLongString(t.skipCommentsHandler)
if err != nil {
	return "", err
}
if wasConsumed {
	return ret.String(), nil
} else {
        // The ' was not part of a long string ending.
        ret.writeByte(byte(c))
}
```

ion/tokenizer.go
```diff
@@ -582,20 +585,27 @@ func (t *tokenizer) readLongString() (string, error) {
 		if err != nil {
 			return "", err
 		}
+		if isInvalidChar(c) {
+			return "", &SyntaxError{"Invalid character", t.pos}
+		}
 
 		switch c {
 		case -1:
 			return "", t.invalidChar(c)
```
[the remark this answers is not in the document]
fernomac: t.invalidChar, as previously suggested, does the right thing here (and should probably have a better name :)).

ion/tokenizer.go
```diff
@@ -1263,3 +1273,24 @@ func (t *tokenizer) unread(c int) {
 	t.pos--
 	t.buffer = append(t.buffer, c)
 }
+
+func isInvalidChar(c int) bool {
+	if c < 0x00 || c > 0x1F {
```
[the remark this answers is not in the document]
fernomac: I found a couple things easier treating EOF as a character instead of an error, and it's an internal-only API so it felt like a good trade-off. It's very possible that's just my old C habits speaking though. :) Feel free to refactor it if you think it improves readability.

ion/tokenizer.go
```diff
@@ -1263,3 +1273,24 @@ func (t *tokenizer) unread(c int) {
 	t.pos--
 	t.buffer = append(t.buffer, c)
 }
+
+func isInvalidChar(c int) bool {
```
fernomac, line 1277: This name's a bit ambiguous at the top level. Maybe `isInvalidStringChar`? It'd fit nicely alongside the other `isXxx` functions in textutils.go.

ion/tokenizer.go
fernomac: It'll also return false if it skipped over a `'''   /* pause in the string */   '''` sequence, in which case you don't want to keep the '. :( This seems correct to me in the short term, `skipEndOfLongString` should probably should be refactored to return a tri-state in the longer term.

ion/tokenizer.go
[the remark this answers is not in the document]
fernomac: :+1: for having this treat -1 as an invalid char. In both places we're calling it we're checking for -1 immediately after, which you could then remove.

ion/tokenizer.go
```diff
@@ -1263,3 +1273,24 @@ func (t *tokenizer) unread(c int) {
...
+	if c < 0x00 || c > 0x1F {
+		return false
+	}
+	if isWhiteSpaceChar(c) || isNewLineChar(c) {
+		return false
+	}
+	return true
+}
+
+func isWhiteSpaceChar(c int) bool {
```
fernomac, line 1287: `isWhitespace` is similarly-named but works a bit differently. Maybe worth combining them? Otherwise maybe rename this one to `isStringWhitespace` or something like that.

ion/tokenizer.go
Armanbqt: Now that's embarrassing! I read [this comment](https://github.com/amzn/ion-go/pull/58#discussion_r436227420) and thought you're suggesting to change the name of the function, totally misread it.
I originally had named it ```isProhibitedControlChar```. Changing it back

ion/tokenizer.go
Armanbqt: Yes, I almost refactored ```skipEndOfLongString``` but then I decided to KISS.
I can create an issue for it.

ion/tokenizer.go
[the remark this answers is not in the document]
Armanbqt: Modified and created #60 to refactor ```read()```

ion/tokenizer.go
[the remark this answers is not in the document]
Armanbqt: Fixed.

ion/tokenizer.go
```diff
@@ -542,11 +542,11 @@ func (t *tokenizer) readString() (string, error) {
 		if err != nil {
 			return "", err
 		}
-
-		switch c {
-		case -1, '\n':
+		if isProhibitedControlChar(c) || c == '\n' {
```
zslayton, line 545: I think we should explicitly handle the `EOF`/`-1` case here and below. `isProhibitedControlChar` will catch it, but the method name makes me think it's only looking for [ASCII control characters](http://ascii-table.com/control-chars.php), which doesn't include -1.

Armanbqt: I see your point, but that means going back to [this comment](https://github.com/amzn/ion-go/pull/58#discussion_r437438308). 
We can either change the function name to something more generic like ```invalidStringCharacter``` or put the logic before [this commit](https://github.com/amzn/ion-go/pull/58/commits/43f7b28c8059f179535f48afc93137c6ef39869e) back.

zslayton: There's a subtle difference here. We've gone from:

```go
if isInvalidChar(c) {
    // Doesn't handle EOF, but could have based on the name
}
switch c {
case -1: // Handles EOF
```
to
```go
if isProhibitedControlChar(c) {
     // Handles EOF, but shouldn't based on the name.
}
```

I think `isProhibitedControlChar` is a more precise/communicative name, so I'd like to keep it. But that means we need the explicit EOF check:
```go
if c == -1 || c == '\n' || isProhibitedControlChar(c) {
    // Handles EOF
}
```

I think this is especially helpful since someone new to the codebase could reasonably expect EOF to be handled by the
```go
if err != nil {
// ...
}
```
above, since `err` is how the standard library reports `EOF`.

Armanbqt: Cool. Updated the pull request. Thanks.

PR #62, opened 2020-06-10
docs: add users section to readme

danielgtaylor:
This small change adds a "users" section to the README that can point at open source projects written in Go that support Amazon Ion by using this library. I've added one of my projects as an example with some text to encourage others to do the same.

Why link to these external projects?

1. They provide a real-world example of using the library that others can learn from
2. They provide tools that may be useful in testing or adopting Amazon Ion
3. They can help create community and a useful feedback mechanism for bugs & features

A little background on my tool: it's a CLI for HTTP APIs that has built-in content negotiation features and supports a bunch of content types out of the box, so for example if the API server sends binary Amazon Ion instead of plain JSON then Restish can parse and display the resulting data.


therapon:
Thanks for the contribution. The description of you tool sounds interesting, I will give it a go :) 

You bring forward a very interesting and valuable point--the addition of clients to the library. 

I'll keep this PR open for a while (if that is OK with you) mainly because we need to figure out the best way to present this information and where. The repo is one option, there is also the Ion web page(s) rooted at http://amzn.github.io/ion-docs/ 

My initial reaction here is that we probably want these type of links on both 

* the web page as a consolidate view helping people decide on language/tools for ion 
* the repo providing people easy access for tools in the language of the library they are already on 

Let me circulate this with the team and upon consensus we can then act.

therapon:
Minor rewording so I made a new PR #126

PR #64, opened 2020-06-12
Moving from json tags to ion tags for Marshal/Unmarshal

### LINKED ISSUE #63: Redefine JSON tag to Ion tag for Marshal/Unmarshal
Given that users will be working with Ion, it will be clearer to use the Ion annotation instead of the json annotation when marshalling and unmarshalling Ion.

However, it is still valuable to make a reference to the json annotation pattern in the readme.

matthewh-BQ:
Resolves #63 

Given that users will be working with Ion, this change is more for clarity and moves away from the json tag to using the ion tag when marshalling and unmarshalling.


fernomac:
Oh wow, I *vividly* remember basing this on a yaml implementation somewhere that just reused the ‘json’ tag, but I just did a check aaaand nope all of the ones I’m familiar with use ‘yaml’. 😅😅😅 :+1: for switching to ‘ion’.

PR #72, opened 2020-06-18
Use go-cmp for comparison in integration_test

Armanbqt:
Currently integration_test is using ```reflect.DeepEqual()``` to check equivalency.
There is at lease one case that we know of where ```DeepEqual()``` is not working as expected and is when float values have ```NaN```. And there might be other cases too.

For now we have a customized ```cmpOption``` for ```FloatType``` to cover the above case, and other types are using the regular ```cmp.Equal()```, **except** for ```DecimalType``` and ```IntType```.

In order to have ```cmp.Equal()``` working for ```DecimalType```, we might need to [explicitly track precision](https://github.com/amzn/ion-go/blob/master/ion/decimal.go#L22).

As for ```IntType``` more investigation is needed, but I probably the issues is with 0 and null.Int cases


Armanbqt:
Closing -- Not ready yet, it was supposed to be a PR into Armanbqt/ion-go fork

PR #74, opened 2020-06-19
Binary ordered structs

Armanbqt:
Ordered structs in binary encoding are presented with ```D1```, and the size of that struct comes in the byte after ```D1```.

### DIFF
--- a/ion/bitstream.go
+++ b/ion/bitstream.go
@@ -173,9 +173,10 @@ func (b *bitstream) Next() error {
 	// Parse the tag.
 	code, len := parseTag(c)
 
-	// Binary-encoded structs with len 1 are special cases of sorted field-names.
+	// Structs with a length code of 1 are a special case. Their length is always encoded
+	// as a VarUInt and their field names appear in ascending symbol ID order.
 	if code == bitcodeStruct && len == 1 {
-		len, err = b.handleOrderedStruct()
+		len, _, err = b.readVarUintLen(b.remaining())
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


ion/bitstream.go
```diff
@@ -949,3 +958,17 @@ func (b *bitstack) pop() {
...
+func (b *bitstream) handleOrderedStruct() (uint64, error) {
+	c, err := b.read()
+	if err != nil {
+		return 0, err
+	}
+	if c == -1 {
+		b.code = bitcodeEOF
+		return 0, nil
+	}
+	length := c & 0x7F
```
zslayton, line 972: Here's the relevant bit of [the `struct` encoding spec](http://amzn.github.io/ion-docs/docs/binary.html#13-struct):

> When L is 1, the struct has at least one symbol/value pair, the length field exists, and the field name integers are sorted in increasing order.

The `length` field is a `VarUInt`, so we have to allow for the case where it takes more than one byte to represent it. We should use the `readVarUintLen` function instead.

ion/bitstream.go
```diff
@@ -172,6 +172,15 @@ func (b *bitstream) Next() error {
 
 	// Parse the tag.
 	code, len := parseTag(c)
+
+	// Binary-encoded structs with len 1 are special cases of sorted field-names.
```
zslayton, line 176: The sort order for field names in this case is a bit un-intuitive. I suggest spelling it out:

```
// Structs with a length code of 1 are a special case. Their length is always encoded as a VarUInt
// and their field names appear in ascending symbol ID order.
```

PR #76, opened 2020-06-19
String with ISO-8859-1 characters

Armanbqt:
This fix is to detect strings with latine-1 (ISO-8859-1) characters.

### DIFF
--- a/ion/bitstream.go
+++ b/ion/bitstream.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"math"
 	"math/big"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


ion/bitstream.go
```diff
@@ -638,7 +639,27 @@ func (b *bitstream) ReadString() (string, error) {
...
 
-	return string(bs), nil
+	return toUtf8String(bs, b.pos)
+}
+
+func toUtf8String(buf []byte, pos uint64) (string, error) {
+	n := len(buf)
+	str := strings.Builder{}
+	re := bytes.NewReader(buf)
+	for n > 0 {
```
zslayton, line 649: I notice that the `utf8` package offers a function called [`Valid`](https://golang.org/pkg/unicode/utf8/#Valid) that can tell you up-front whether the provided slice contains valid UTF8. Using that will likely be faster than copying a rune at a time to a string builder; if `Valid(buf)` is true, you can return `string(buf)`.

Armanbqt: Thanks Zack!
It simplified the PR a lot!

PR #77, opened 2020-06-19
Fix invalid negative int zero

Armanbqt:
Ion specs for [integer](http://amzn.github.io/ion-docs/docs/binary.html#2-and-3-int) reads:

> Zero is always stored as positive; negative zero is illegal.

### DIFF
--- a/ion/bitstream.go
+++ b/ion/bitstream.go
@@ -407,18 +407,21 @@ func (b *bitstream) ReadInt() (interface{}, error) {
 	}
 
 	var ret interface{}
+	isZero := false
 	switch {
 	case b.len == 0:
 		// Special case for zero.
 		ret = int64(0)
+		isZero = true
 
-	case b.len < 8, (b.len == 8 && bs[0]&0x80 == 0):
+	case b.len < 8, b.len == 8 && bs[0]&0x80 == 0:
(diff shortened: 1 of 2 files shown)


ion/bitstream.go
```diff
@@ -427,12 +430,18 @@ func (b *bitstream) ReadInt() (interface{}, error) {
 	default:
 		// Need to go big.Int.
 		i := new(big.Int).SetBytes(bs)
+		isZero = i.Cmp(big.NewInt(0)) == 0
```
zslayton, line 433: Constructing a bigint can be expensive as it often requires a heap allocation; is there a constant for `0` that we could re-use instead? If not, can we define one in this package?

Armanbqt: I couldn't find any constant value for 0, that's why I constructed bigInt 0.

Seems like big.Int is an Invalid constant type.
bigInt

I used [```BitLen()```](https://golang.org/pkg/math/big/#Int.BitLen) to check for zero ([Example in play.golang](https://play.golang.org/p/RAa7nzSgHyj))

zslayton: > I used BitLen() to check for zero

Clever workaround, I like it. :+1:

Issue #71, opened 2020-06-15
Add IsInStruct logic to ion reader/writer to match ion-dotnet


justing-bq:
Currently in ion-hash-go, the hash reader/writer have isInStruct() methods defined that just check if the current type is ion.StructType.

However in dotnet, the IsInStruct logic is defined for each ion reader/writer:

RawBinaryReader:
public bool IsInStruct => this.CurrentDepth > 0 && this.parent != null && this.parent.Type() == IonType.Struct;

RawTextReader: 
public bool IsInStruct => this.containerStack.Count > 0 && this.containerStack.Peek() == IonType.Struct;

SystemTreeReader:
public bool IsInStruct => this.CurrentDepth > 0 && this.parent != null && this.parent.Type() == IonType.Struct;

IonTextWriter:
public override bool IsInStruct => this.isInStruct;

IonTreeWriter:
public override bool IsInStruct => ((IIonValue)this.currentContainer).Type() == IonType.Struct;

ManagedBinaryWriter:
public override bool IsInStruct => this.userWriter.IsInStruct;

This difference in implementation accounts for some of the failing unit tests in ion-hash-go.

PR #85, opened 2020-06-23
Add InStruct() to Reader and Writer API

Armanbqt:
InStruct() needs to be in the API so IonHash can use it.


Armanbqt:
This is to address #71 
Please let me know if there are any issues with changes of the API.

cmd/ion-go/eventwriter.go
```diff
@@ -194,11 +195,13 @@ func (e *eventwriter) BeginStruct() error {
 		return err
 	}
 	e.depth++
+	e.inStruct = true
 	return nil
 }
 
 func (e *eventwriter) EndStruct() error {
 	e.depth--
+	e.inStruct = false
```
fernomac, line 204: `{structs:{can:"end"},inside:{other:"structs"}}` :(

ion/textreader_test.go
```diff
@@ -423,6 +423,38 @@ func TestTrsToString(t *testing.T) {
 	}
 }
 
+func TestInStruct(t *testing.T) {
+	r := NewReaderStr("[ { a:() } ]")
+
+	r.Next()
+	r.StepIn() // In the list, before the struct
+	if r.InStruct() {
+		t.FailNow()
```
fernomac, line 432: `t.Fatal("InStruct returned true before we were in a struct")`?

ion/reader.go
```diff
@@ -13,7 +13,7 @@ import (
 // A Reader reads a stream of Ion values.
 //
 // The Reader has a logical position within the stream of values, influencing the
-// values returnedd from its methods. Initially, the Reader is positioned before the
+// values returned from its methods. Initially, the Reader is positioned before the
```
fernomac, line 16: Thanks! :)

cmd/ion-go/eventwriter.go
Armanbqt: Good catch. I'll fix it.

ion/textreader_test.go
Armanbqt: Done.

ion/reader.go
```diff
@@ -148,6 +148,9 @@ type Reader interface {
 	// ByteValue returns the current value as a byte slice (if that makes sense). It returns
 	// an error if the current value is not an Ion clob or an Ion blob.
 	ByteValue() ([]byte, error)
+
+	// InStruct indicates if the reader is currently positioned in a struct.
+	InStruct() bool
```
zslayton, line 153: Matter of taste: I'd prefer the name `IsInStruct` to make it obvious that this returns a `bool`.

cmd/ion-go/eventwriter.go
```diff
@@ -15,6 +15,7 @@ type eventwriter struct {
 	depth       int
 	fieldname   string
 	annotations []string
+	inStruct    map[int]bool
```
zslayton, line 18: Could we re-use the `ctxstack` concept that's leveraged by the reader? That would avoid heap allocations.

Armanbqt: ```ctxstack``` is internal right now and can be only used within ```ion``` directory.
Do we want to make ```ctxstack``` public?

Armanbqt: We can move ```ctx.go``` to 'internal' directory in order to be able use that in cmd package. But then we will have cycle imports (as ctx is using ion.Type).
A similar case is under discussion/investigation in an [ion-hash-go pull request](https://github.com/amzn/ion-hash-go/pull/31).

ion/reader.go
Armanbqt: Renamed to ```IsInStruct```

cmd/ion-go/eventwriter.go
Armanbqt: To unlock Ion-hash, can we merge this pr and revisit this concern in #91 ?

PR #86, opened 2020-06-30
Fix for empty lobs

### LINKED ISSUE #82: Malformed blobs
Malformed blobs should not form valid readers. An example is in [blobLenTooLarge.10n](https://github.com/amzn/ion-go/blob/master/ion/integration_test.go#L147) test file.

Armanbqt:
Resolves #82 

When we are reading an empty Lob, the ```len``` is 0 and it causes ```b.readN(b.len)``` to return ```nil```.


ion/bitstream.go
```diff
@@ -660,6 +660,11 @@ func (b *bitstream) ReadBytes() ([]byte, error) {
 		panic("not a lob")
 	}
 
+	// A0 and 90 are special cases, denoting an empty blob and an empty clob respectively, with b.len == 0.
+	if b.len == 0 {
+		return make([]byte, 0), nil
```
zslayton, line 665: ~~Rather than instantiating an empty byte array each time, could we return a constant?~~

Go doesn't allow `const` slices. :)

It looks like [we can avoid a small amount of overhead](https://stackoverflow.com/q/29164375) by using an empty slice literal instead of calling `make`, however.

```suggestion
		return []byte{}, nil
```

ion/bitstream.go
```diff
@@ -660,6 +660,11 @@ func (b *bitstream) ReadBytes() ([]byte, error) {
 		panic("not a lob")
 	}
 
+	// A0 and 90 are special cases, denoting an empty blob and an empty clob respectively, with b.len == 0.
+	if b.len == 0 {
+		return []byte{}, nil
```
fernomac, line 665: This misses the update to `b.state` down below; that seems like it might be a bug?

Armanbqt: Good point, I'll address it.
Thanks!

Armanbqt: PR #87

PR #87, opened 2020-06-30
Update state when reading empty lob issue

Armanbqt:
To address [this comment](https://github.com/amzn/ion-go/pull/86#discussion_r447920472).

### DIFF
--- a/ion/bitstream.go
+++ b/ion/bitstream.go
@@ -660,14 +660,16 @@ func (b *bitstream) ReadBytes() ([]byte, error) {
 		panic("not a lob")
 	}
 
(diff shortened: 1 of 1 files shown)


ion/bitstream.go
```diff
@@ -660,14 +660,14 @@ func (b *bitstream) ReadBytes() ([]byte, error) {
 		panic("not a lob")
 	}
 
+	bs := []byte{}
```
zslayton, line 663: I imagine this still involves a heap allocation; that would be wasted effort if `b.len > 0`. Can you make this an `else` block on the `if` below?

PR #89, opened 2020-07-02
Add timestamp precision rounding

byronlin-bq:
Currently timestamps are truncated at nanoseconds. Add rounding up if value is 5 or greater.

Eg. 
```
"1234-05-06T07:08:09.0001001004Z" -> "1234-05-06T07:08:09.000100100Z"
"1234-05-06T07:08:09.0001001005Z" -> "1234-05-06T07:08:09.000100101Z"
```

Issue #, if available:

Description of changes:


ion/bitstream.go
```diff
@@ -541,41 +541,55 @@ func (b *bitstream) ReadTimestamp() (time.Time, error) {
...
 
+	if overflow {
+		date = date.Add(time.Second)
+	}
 	return date.In(time.FixedZone("fixed", int(offset)*60)), nil
 }
 
-// ReadNsecs reads the fraction part of a timestamp and truncates it to nanoseconds.
-func (b *bitstream) readNsecs(len uint64) (int, error) {
+// ReadNsecs reads the fraction part of a timestamp and rounds to nanoseconds.
```
zslayton, line 568: Please add a sentence to this comment explaining the new `(int, bool, error)` return type.

ion/bitstream_test.go
```diff
@@ -109,3 +112,188 @@ func TestBitcodeString(t *testing.T) {
 		}
 	}
 }
+
+func TestBinaryReadTimestamp(t *testing.T) {
+	test := func(ion []byte, eval string) {
```
zslayton, line 117: I think `eval` is intended to mean `expectedValue`, but the word `eval` is widely used to mean `evaluate` or `evaluation`. Could you expand it?

ion/bitstream_test.go
```diff
@@ -109,3 +112,188 @@ func TestBitcodeString(t *testing.T) {
...
+			b := bitstream{}
+			b.InitBytes(ion)
+			b.Next()
+
+			val, err := b.ReadTimestamp()
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			et, err := time.Parse(time.RFC3339Nano, eval)
```
zslayton, line 129: Similar comment here: if `et` is meant to mean `expectedTime`, please expand it.

ion/decimal.go
```diff
@@ -231,21 +231,29 @@ func (d *Decimal) upscale(scale int32) *Decimal {
 	}
 }
 
-// Trunc attempts to truncate this decimal to an int64, dropping any fractional bits.
-func (d *Decimal) Trunc() (int64, error) {
+func (d *Decimal) checkToUpscale() (*Decimal, error) {
```
zslayton, line 234: Please add a comment to this method explaining what it means to "upscale" a `Decimal`.

ion/decimal.go
```diff
@@ -231,21 +231,29 @@ func (d *Decimal) upscale(scale int32) *Decimal {
...
+// Trunc attempts to truncate this decimal to an int64, dropping any fractional bits.
+func (d *Decimal) Trunc() (int64, error) {
+	ud, err := d.checkToUpscale()
+	if err != nil {
+		return 0, err
+	}
+	d = ud
 	str := d.n.String()
 
 	want := len(str) - int(d.scale)
```
zslayton, line 259: I don't think I understand what `want` is meant to convey here. The variable contains the index of the first rune in the `Decimal`'s `string` representation that we don't want to keep. Maybe `truncateTo`?

ion/decimal.go
```diff
@@ -256,6 +264,19 @@ func (d *Decimal) Trunc() (int64, error) {
 	return strconv.ParseInt(str[:want], 10, 64)
 }
 
+// Round attempts to truncate this decimal to an int64, rounding any fractional bits.
+func (d *Decimal) Round() (int64, error) {
```
zslayton, line 268: Based on the name `Round`, I would expect this method to return the nearest integer value as a`Decimal`. (This is analogous to the behavior of the `Truncate` method below, which returns a `Decimal`.) Truncating the value to fit in the `int64` range would be pretty surprising, as that could lead to substantial data loss. What's the rationale behind changing it to an `int64`?

ion/bitstream_test.go
fernomac: I used an `e` prefix to mean "expected" throughout the test code, including a lot of `eval`s for "expected value", so this one is presumably my fault. :) I prefer the brevity, especially for locally-scoped variables where I don't have to go far to see how they get used, but you're not wrong and if you want to expand them go for it.

ion/decimal.go
byronlin-bq: This function is analogous to `Trunc` function which returns `int64`. `int64` will be sufficient because `readNsecs` function will shift decimal to 9 decimal places to the left before calling this function.

`func (b *bitstream) readNsecs(len uint64) (int, bool, error) {
	d, err := b.readDecimal(len)
	if err != nil {
		return 0, false, err
	}

	nsec, err := d.ShiftL(9).Trunc()
	if err != nil || nsec < 0 || nsec > 999999999 {
		msg := fmt.Sprintf("invalid timestamp fraction: %v", d)
		return 0, false, &SyntaxError{msg, b.pos}
	}

	nsec, err = d.ShiftL(9).Round()
	if err != nil {
		msg := fmt.Sprintf("invalid timestamp fraction: %v", d)
		return 0, false, &SyntaxError{msg, b.pos}
	}

	// Overflow to second.
	if nsec == 1000000000 {
		return 0, true, nil
	}

	return int(nsec), false, nil
}`

ion/bitstream_test.go
byronlin-bq: Updated to `expectedTime`

ion/bitstream.go
byronlin-bq: Updated.

ion/bitstream_test.go
byronlin-bq: Updated to `expectedValue`

ion/decimal.go
byronlin-bq: Added comment.

ion/decimal.go
byronlin-bq: Updated to `truncatetTo`

ion/decimal.go
zslayton: > This function is analogous to Trunc function which returns int64

I hadn't noticed that; I would also change `Trunc` 😛 .

The problem is that while `int64` is suitable for *this* use case, we're making these functions part of `Decimal`'s public interface. Users working with a `Decimal` representation of their data will have varying requirements.

I would suggest either:
* making `Trunc` and `Round` private functions and change their names to reflect the data type conversion.
* making `Trunc` and `Round` return a `Decimal` (like `Truncate` does) and offering a separate method to convert to int64. This method would check for overflow and make sure the `Decimal` was an integer.

The latter is obviously more work; I'm fine with taking them private.

byronlin-bq: Updated to private functions

PR #92, opened 2020-07-03
Using go-cmp for equality check on timestamp, floats and decimal

Armanbqt:
To get a more accurate and custom way of checking equality between Ion values, we are going to use ```go-cmp``` package.

The implementation is limited to Decimal, Float and Timestamp, as we need custom equivalency. Other types are using ```reflect.DeepEqual()```.

-----
There are some discussions in [this draft pull request](https://github.com/Armanbqt/ion-go/pull/9) on my fork; adding it here to keep the history.


ion/cmp_test.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func (thisTimestamp ionTimestamp) eq(other ionEqual) bool {
+	if val, ok := other.(ionTimestamp); ok {
+		return thisTimestamp.Time.Equal(val.Time)
+	}
+	return false
+}
+
+func cmpAnnotations(thisAnnotations, otherAnnotations []string) bool {
```
zslayton, line 40: Since this is comparing `string` slices, I think you might be able to `return DeepEquals(thisAnnotations, otherAnnotations)`?

ion/cmp_test.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	res := false
+	for idx, this := range thisValues {
+		other := otherValues[idx]
+		thisType, otherType := reflect.TypeOf(this), reflect.TypeOf(other)
+
+		if thisType != otherType {
+			return false
+		}
```
zslayton, line 72: I think you can use your `haveSameTypes` helper method for this.

ion/cmp_test.go
```diff
@@ -0,0 +1,145 @@
...
+		thisTimestamp := ionTimestamp{val}
+		return thisTimestamp.eq(ionTimestamp{otherValue.(time.Time)})
+	default:
+		return false
+	}
+}
+
+func strNullTypeCmp(this, other interface{}) bool {
+	thisStr := this.(string)
+	otherStr := other.(string)
```
zslayton, line 139: Will this always succeed? `other` isn't guaranteed to be a `string`. (Or maybe it just turns into a `nil`, which is ok for passing to `cmp.Equal`?)

ion/cmp_test.go
```diff
@@ -0,0 +1,145 @@
...
+	for idx, this := range thisValues {
+		other := otherValues[idx]
+		thisType, otherType := reflect.TypeOf(this), reflect.TypeOf(other)
+
+		if thisType != otherType {
+			return false
+		}
+
+		switch this.(type) {
+		case string: // null.Sexp, null.List, null.Struct
```
zslayton, line 75: Nit: the typed null syntax is always lowercase (`null.sexp`, `null.list`, etc)

ion/cmp_test.go
```diff
@@ -0,0 +1,145 @@
...
+	return res
+}
+
+func cmpFloats(thisValue, otherValue interface{}) bool {
+	if !haveSameTypes(thisValue, otherValue) {
+		return false
+	}
+
+	switch val := thisValue.(type) {
+	case string: // null.float
```
zslayton, line 95: Can you point me to where the string representations of typed nulls are being created?

ion/cmp_test.go
Armanbqt: This is called in ```switch case```s, only when  ```case string``` (lines 76 - 96 - 112 - 128)

ion/cmp_test.go
Armanbqt: In ```integration_test``` we replace typed nulls with their string representations: [readCurrentValue()](https://github.com/amzn/ion-go/blob/master/ion/integration_test.go#L721)

ion/cmp_test.go
Armanbqt: Thanks

ion/cmp_test.go
Armanbqt: True

ion/cmp_test.go
Armanbqt: Done

ion/integration_test.go
```diff
@@ -46,6 +47,9 @@ func (i *ionItem) equal(o ionItem) bool {
 	if !cmpAnnotations(i.annotations, o.annotations) {
 		return false
 	}
+	if !cmpFieldNames(i.fieldName, o.fieldName) {
```
zslayton, line 50: This test requires that the two structs' field names be in the same order, but the Ion spec allows field names to be in any order when checking for equality.

ion/integration_test.go
```diff
@@ -732,6 +735,11 @@ func readCurrentValue(t *testing.T, reader Reader) ionItem {
 		ionItem.annotations = an
 	}
 
+	fn := reader.FieldName()
+	if fn != "" {
```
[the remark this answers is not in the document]
Armanbqt: Right, I merged master to this PR

ion/cmp_test.go
```diff
@@ -134,12 +93,115 @@ func cmpTimestamps(thisValue, otherValue interface{}) bool {
...
+		res = containersEquality(thisItem, otherItem)
+
+		if !res {
+			return false
+		}
+	}
+	return res
+}
+
+func cmpStruct(thisValues, otherValues []interface{}) bool {
```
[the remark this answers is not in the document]
Armanbqt: For a case like [this test](https://github.com/amzn/ion-tests/blob/master/iontestdata/good/non-equivs/structs.ion#L7), I assume we need to check both values and field name.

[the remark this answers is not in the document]
zslayton: I was asking because I couldn't see where the field name equality was happening.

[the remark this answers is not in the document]
Armanbqt: In ```integration_test``` line 50 (where you already asked another question and I was checking it now ;) )

ion/integration_test.go
Armanbqt: So probably I have to move this to ```cmpStruct``` function in ```cmp_test``` file. We know we have field name only when we are checking structs

ion/cmp_test.go
```diff
@@ -0,0 +1,205 @@
...
+func containersEquality(this, other interface{}) bool {
+	switch this.(type) {
+	case string: // null.list, null.sexp, null.struct
+		if strNullTypeCmp(this, other) {
+			return true
+		}
+	default:
+		otherItem := other.(ionItem)
+		thisItem := this.(ionItem)
+		if thisItem.fieldName == otherItem.fieldName && thisItem.equal(otherItem) {
```
Armanbqt, line 200: ```thisItem.fieldName == otherItem.fieldName``` can only be done when both values are converted to ```ionItem``` so I put this here rather than inside ```cmpStruct```.
This still can avoid execution of ```.equal()``` if inferred as false

PR #93, opened 2020-07-07
Fix issues with struct

Armanbqt:
1- Empty string is a valid value for field names.
Default value for strings in Go is empty string. Checking whether a fieldName is a set or not based on checking them against "", disallow us for having something like ```{"":empty}```.
One way to solve this issue is to change ```FieldName() string``` in Reader.go to ```FieldName() *string``` and set ```nil``` as ```reader.fieldName```'s default value in both ```textReader``` and ```binaryReader```. Then we will be able to change check if a fieldName is not set by ```r.fieldName != nil```. However, I do not think using nil as a magic value is a good idea in Go, and secondly changing return type of ```FieldName()``` is changing the API.
Alternatively in [this commit](https://github.com/amzn/ion-go/commit/d6c914ba3c51e8e144feddb6c4ab4104278701df), I introduced a constant representing a value for when field name is not set (for now I chose UNDEFINED, but open to other suggestions if we decided to go with this solution). The downside is we are omitting one certain word to be a field name and also it is not aligned with other implementations of Ion libraries. 

2- If a struct is marked as an ordered struct, there must be at least one symbol/value pair: [this commit](https://github.com/amzn/ion-go/commit/33d9fc1d70043d5394f6df86ea9d3d3128624c08)

### DIFF
--- a/ion/reader.go
+++ b/ion/reader.go
@@ -384,7 +384,7 @@ func (r *reader) ByteValue() ([]byte, error) {
 
 // Clear clears the current value from the reader.
 func (r *reader) clear() {
-	r.fieldName = ""
+	r.fieldName = undefinedFieldName
 	r.annotations = nil
 	r.valueType = NoType
 	r.value = nil
--- a/ion/reader.go
+++ b/ion/reader.go
@@ -85,7 +85,7 @@ type Reader interface {
 	// FieldName returns the field name associated with the current value. It returns
 	// the empty string if there is no current value or the current value has no field
 	// name.
-	FieldName() string
+	FieldName() *string
 
 	// Annotations returns the set of annotations associated with the current value.
 	// It returns nil if there is no current value or the current value has no annotations.
@@ -187,7 +187,7 @@ type reader struct {
 	eof bool
 	err error
 
-	fieldName   string
+	fieldName   *string
 	annotations []string
 	valueType   Type
 	value       interface{}
@@ -209,7 +209,7 @@ func (r *reader) IsNull() bool {
 }
 
 // FieldName returns the current value's field name.
-func (r *reader) FieldName() string {
+func (r *reader) FieldName() *string {
 	return r.fieldName
 }
 
@@ -384,7 +384,7 @@ func (r *reader) ByteValue() ([]byte, error) {
 
 // Clear clears the current value from the reader.
 func (r *reader) clear() {
-	r.fieldName = ""
+	r.fieldName = nil
 	r.annotations = nil
 	r.valueType = NoType
 	r.value = nil
--- a/ion/reader.go
+++ b/ion/reader.go
@@ -384,7 +384,7 @@ func (r *reader) ByteValue() ([]byte, error) {
 
 // Clear clears the current value from the reader.
 func (r *reader) clear() {
-	r.fieldName = undefinedFieldName
+	r.fieldName = ""
 	r.annotations = nil
 	r.valueType = NoType
 	r.value = nil
--- a/cmd/ion-go/eventwriter.go
+++ b/cmd/ion-go/eventwriter.go
@@ -253,11 +253,11 @@ func clobify(val []byte) string {
 
 func (e *eventwriter) write(ev event) error {
 	name := e.fieldname
-	e.fieldname = "UNDEFINED"
+	e.fieldname = ""
 	annos := e.annotations
 	e.annotations = nil
 
-	if name != "UNDEFINED" {
+	if name != "" {
 		ev.FieldName = &token{Text: name}
 	}
 
--- a/cmd/ion-go/eventwriter.go
+++ b/cmd/ion-go/eventwriter.go
@@ -13,7 +13,7 @@ type eventwriter struct {
 	enc *ion.Encoder
 
 	depth       int
-	fieldname   string
+	fieldname   *string
 	annotations []string
 	inStruct    map[int]bool
 }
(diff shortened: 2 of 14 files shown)


fernomac:
Hm, I don't love the idea of having a secret, non-empty-string value that can't get used as a field name. How would you feel about the approach [here](https://github.com/amzn/ion-go/compare/master...fernomac:empty-fieldnames), using `*string` internally, keeping the interface backwards compatible, and adding a `Reader.HasFieldName` for the (rare?) case where someone needs to explicitly check if an empty string from `Reader.FieldName` means no field name or the actual empty string?

Armanbqt:
Discussed offline and it was decided to change ```FieldName() string``` in Reader interface to ```FieldName() *string```, and consequently ```filedName ``` attribute of ```reader``` struct.

ion/reader.go
```diff
@@ -85,7 +85,7 @@ type Reader interface {
 	// FieldName returns the field name associated with the current value. It returns
 	// the empty string if there is no current value or the current value has no field
 	// name.
-	FieldName() string
+	FieldName() *string
```
fernomac, line 88: Returning a `*string` via the public interface feels pretty awkward an un-go-like to me, especially given that it's going to be non-nil in basically all cases where someone might call it. :( As I suggested previously, I think the interface would be better with this left as a string, and an auxiliary `HasFieldName() bool` method for the rare case where someone is uncertain about whether an empty-string return value means "not in a struct" or "whoever wrote this struct is rude." Thoughts?

If you are going to proceed with this change despite my objections, which is your right, the comment should be updated to match. :)

zslayton: > ...and an auxiliary HasFieldName() bool method...

From what I can tell, `HasFieldName()` is essentially a synonym for `IsInStruct()`. Is that right?

> Returning a *string via the public interface feels pretty awkward an un-go-like to me

Could you elaborate on this? I liked the idea of returning a `*string` because it makes it very clear that there might not be a field name.

This:
```go
if reader.IsInStruct() {
   fieldName := *reader.FieldName()
}
```
feels very similar to this:
```go
if reader.HasFieldName() {
  fieldName := reader.FieldName()
}
```

On the other hand, detecting an absent field name after-the-fact is simpler with a pointer. Compare this:
```go
fieldName := reader.FieldName()
if fieldName == "" && !reader.HasFieldName() {
  // There wasn't a field name
}
```

with this:
```go
fieldName := reader.FieldName()
if fieldName == nil {
  // There wasn't a field name
}
```

Using a fieldName that's `nil` will lead to an error instead of silently proceeding with a surprising value. Using a `string` does let you get away with simply calling `FieldName()` whether it's sensible or not, but I'm unsure that's something we want to encourage.

fernomac: Pedantically speaking, `HasFieldName()` is `IsInStruct() && Type() != NoType`; it will return false immediately after a call to `StepIn()`, for example. :)

Pointers to basic types like strings feel weird because they don't really get used much in idiomatic go (e.g. see all the complaints [here](https://github.com/aws/aws-sdk-go-v2/issues/205) and in linked issues, although in that case I'm not sure there's a better solution). In particular: syntax sugar for pointers-to-structs means it's very uncommon to ever have to manually dereference a pointer; doing so feels strange and a bit dangerous, the kind of thing you're only supposed to do if you're interfacing with a C library.

If you think it's common for folks to check if `FieldName()` has in fact returned a field name, a more idiomatic approach would be `FieldName() (string, bool)`, in the style of go built-in [map type](https://blog.golang.org/maps). My concern with that approach here is that, in almost all cases, I expect the caller already knows to expect a field name by virtue of being inside of a struct (e.g. all three references to FieldName in `unmarshal.go`). It feels unfortunately to make them all explicitly ignore the boolean that they know at compile time will always be true.

```go
// can't do write(in.FieldName())
fn, _ := in.FieldName()
write(fn)
```

In the relatively niche case where someone's handling a sequence of values that may or may not have field names at runtime (as in [process.go](https://github.com/amzn/ion-go/blob/master/cmd/ion-go/process.go#L178)), it seems pretty clear to do:

```go
if in.HasFieldName() {
  out.FieldName(in.FieldName())
}
```

vs

```go
fn, ok := in.FieldName()
if ok {
  out.FieldName(fn)
}
```

(In practice, I think it's probably fine if you feel differently, I imagine most folks who are concerned about idiomatic go will be using Unmarshal not the underlying Reader. :))

therapon: A field name that is the empty string `""` is *valid* per the Ion Spec. http://amzn.github.io/ion-docs/docs/spec.html#struct. 

So there is a case where we can be inside a struct and the empty string is a valid value for a field name. So there is going to be a test in the client code to check what `""` means.  This means that whether we move to `string` or `*string` there is still a test that needs to be performed by the client if they are to validate correct behaviour here. 

I also agree with the comment that "this is the low level API" and as such there is an expectation that the client code needs to know of these cases. 

I vote for a 1-1 mapping of value to case here 

* `nil` -- something is wrong go figure out what by checking the reader's context 
* non-nil -- all is well you can proceed 

We should make this **explicit** and explain it in the comment for the method. 

Another argument here (though maybe weak) is that in cases where the API is misused and the client does not check the value, a `nil` will error while an `""` can result in data that is valid Ion but not the intended Ion and detecting that you are generating valid Ion but invalid for the use case is harder and riskier in my opinion. 

My vote is to keep the pointer type and add a comment to the API that explicitly spells out the distinction between `nil` and non-`nil` values so that clients are aware of the methods expectations and guarantees.

PR #94, opened 2020-07-08
Support for System Symbols as Annotations in Text Writer

matthewh-BQ:
Text Writer and Text Reader have a concept of SymbolTable.
This introduces the concept into the Text Writer first by bringing down
the SymbolTable to the base writer. When a writer is constructed,
a default system table is created and then referenced to resolve
any Symbols as annotations.

### DIFF
--- a/ion/textwriter.go
+++ b/ion/textwriter.go
@@ -39,7 +39,10 @@ func NewTextWriter(out io.Writer) Writer {
 // NewTextWriterOpts returns a new text writer with the given options.
 func NewTextWriterOpts(out io.Writer, opts TextWriterOpts) Writer {
 	return &textWriter{
-		writer:      writer{out: out},
+		writer:      writer{
+			out: out,
+			lst: NewSymbolTableBuilder().Build(),
+		},
 		opts:        opts,
 		emptyStream: true,
 	}
@@ -314,7 +317,7 @@ func (w *textWriter) beginValue(api string) error {
 	}
 
 	if len(w.annotations) > 0 {
-		if err := w.writeAnnotations(); err != nil {
+		if err := w.writeAnnotations(api); err != nil {
 			return err
 		}
 	}
@@ -372,14 +375,28 @@ func (w *textWriter) writeFieldName(api string) error {
 }
 
 // writeAnnotations writes out the annotations for a value.
-func (w *textWriter) writeAnnotations() error {
+func (w *textWriter) writeAnnotations(api string) error {
 	as := w.annotations
 	w.annotations = nil
 
-	for _, a := range as {
-		if err := writeSymbol(a, w.out); err != nil {
+	// If possible, resolve symbols as their names, else default to ID.
+	symbols := make([]string, len(as))
+	for i, a := range as {
+		id, _ := w.resolve(api, a)
+		name, hasName := w.lst.FindByID(id)
+
+		if hasName {
+			symbols[i] = name
+		} else {
+			symbols[i] = a
+		}
+	}
+
+	for _, s := range symbols {
+		if err := writeSymbol(s, w.out); err != nil {
 			return err
 		}
+
 		if err := writeRawString("::", w.out); err != nil {
 			return err
 		}
--- a/ion/writer.go
+++ b/ion/writer.go
@@ -2,8 +2,11 @@ package ion
 
 import (
 	"errors"
+	"fmt"
 	"io"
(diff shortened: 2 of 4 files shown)


ion/writer.go
```diff
@@ -166,3 +174,24 @@ func (w *writer) clear() {
 	w.fieldName = ""
 	w.annotations = nil
 }
+
+// Resolve resolves a symbol to its ID.
+func (w *writer) resolve(api, sym string) (uint64, error) {
```
zslayton, line 179: In the context of symbols, we usually use the verb "resolve" to mean the inverse: given some symbol ID, look up the corresponding text. Could you rename this? (`findOrCreateSid`, maybe?)

zslayton, line 179: Could you add a comment to the function signature that explains what the `api` variable is for?

ion/textwriter.go
```diff
@@ -372,14 +375,28 @@ func (w *textWriter) writeFieldName(api string) error {
...
 	as := w.annotations
 	w.annotations = nil
 
-	for _, a := range as {
-		if err := writeSymbol(a, w.out); err != nil {
+	// If possible, resolve symbols as their names, else default to ID.
+	symbols := make([]string, len(as))
+	for i, a := range as {
+		id, _ := w.resolve(api, a)
+		name, hasName := w.lst.FindByID(id)
```
zslayton, line 386: I think I'm missing something here:

* We get the SID for the string `a` ("annotation"?)
* If it exists, we look it up in the table and are guaranteed to get back the same string value as `a`
* If it doesn't exist, we fall back to `a`.

It seems like we could skip the whole thing and just write `a`. What value is the symbol lookup getting us?

matthewh-BQ:
@therapon : As discussed earlier. I think there was some clarification on whether this is an issue or not. Current behaviour keeps system symbols annotations as $1, $2, etc. This is currently failing on the TextRoundTrip test because the text writer output differs from the binary writer output. The Binary Writer performs the resolution, so there is a mismatch.

therapon:
> @therapon : As discussed earlier. I think there was some clarification on whether this is an issue or not. Current behaviour keeps system symbols annotations as $1, $2, etc. This is currently failing on the TextRoundTrip test because the text writer output differs from the binary writer output. The Binary Writer performs the resolution, so there is a mismatch.

@matthewh-BQ  I looked at the Ion Java library and I think I know what's tripping us up here. We need to clarify in the API the following two cases when adding an annotation 

1. Add an annotation and the value to add is a string. The intent here (I think) is "here is the text given to you as a `string` and I want this text to be added to the list of annotations." 
1. Add an annotation and the value to add is a Symbol ID (SID). The intent here (I think) is "here is the symbol token, trust me I know what I am doing and how symbol tables work". 

This leads to 2 methods/function one for each case. 

The first case is the one we expect to be used more commonly, take this string and add it to the list of annotations for the current element in the writer. The input is a `string` and therefore the code will take the string and **create a new symbol token with a new SID in the symbol table**

Using Ion Java's [IonWriter#addTypeAnnotation](https://javadoc.io/static/com.amazon.ion/ion-java/1.6.1/com/amazon/ion/IonWriter.html#addTypeAnnotation-java.lang.String-) and passing as argument `"$1"` the Ion Java implementation creates a new symbol token for the string `"$1"` and writes the annotation to the output as  `'$1':: .... `

* See [IonWriterSystem#addTypeAnnotation](https://github.com/amzn/ion-java/blob/master/src/com/amazon/ion/impl/IonWriterSystem.java#L587)
* See [IonWriterSystem#setTypeAnnotations](https://github.com/amzn/ion-java/blob/master/src/com/amazon/ion/impl/IonWriterSystem.java#L576) -- this is a different API since it takes a list and replaces any annotations, this is more these are the annotations period. 


The second case, which requires clients to know what symbol tokens are and how they work in Ion, takes as input a **symbol token** which is used as is and will be resolved with the current symbol table.  

* See [IonWriterSystem#setTypeAnnotationSymbols](https://github.com/amzn/ion-java/blob/master/src/com/amazon/ion/impl/IonWriterSystem.java#L534)


My proposal at this point is to create the API for the 1st use case (here is a string with the text that I wand added as an annotation to the list of annotation's  of this value). The second use case I think we can address later.

matthewh-BQ:
Closing draft PR as per discussion. SymbolTokens and SymbolTables will address the second case. The first case is already supported.

PR #98, opened 2020-07-17
Use Timestamp struct instead of time.Time

justing-bq:


ion/bitstream_test.go
```diff
@@ -253,7 +253,7 @@ func TestBinaryReadTimestamp(t *testing.T) {
 		0x80, // second: 0
 		0xCA, // 10 precision units
 		0x2C, // 44
-	}, "2000-01-01T00:00:00.000000004Z", Nanosecond,  UTC)
+	}, "2000-01-01T00:00:00.000000004Z", Nanosecond, UTC)
```
fernomac, line 256: (An editor that automatically runs `go fmt` before save has made my life a lot more pleasant, may be worth considering. :) I've been quite happy with VS Code's go mode if you're looking for suggestions.)

justing-bq: My vscode is set up to automatically run `go fmt` upon save but my goland requires a 4-key-shortcut to run `go fmt` and I don't always remember to do it as you can see here.

mrutter-amzn: In the preferences in GoLand go to Tools -> File Watchers.  You can add one that runs `go fmt`, or better yet `goimports` on save.

justing-bq: Thanks for the tip. Hopefully this means I'll never have another 'satisfy formatter' commit again.

cmd/ion-go/main.go
```diff
@@ -65,7 +64,7 @@ func printVersion() {
 		w.FieldName("version")
 		w.WriteString(internal.GitCommit)
 
-		buildtime, err := time.Parse(time.RFC3339, internal.BuildTime)
+		buildtime, err := ion.NewTimestampFromStr(internal.BuildTime, ion.Second, ion.UTC)
```
zslayton, line 67: Can we namespace the timestamp-related values more narrowly somehow? `ion.Second` and `ion.UTC` feel out of place at the top-level of an `ion` package to me.

ion/binarywriter_test.go
```diff
@@ -153,22 +153,22 @@ func TestWriteBinarySymbol(t *testing.T) {
 func TestWriteBinaryTimestamp(t *testing.T) {
 	eval := []byte{
 		0x67, 0x80, 0x81, 0x81, 0x81, 0x80, 0x80, 0x80, // 0001-01-01T00:00:00Z
-		0x6E, 0x8E, // 0x0E-bit timestamp
+		0x6D,       // 0x0D-bit timestamp
```
zslayton, line 156:
```suggestion
		0x6D,       // 0x0D-byte timestamp
```

ion/binarywriter_test.go
```diff
@@ -153,22 +153,22 @@ func TestWriteBinarySymbol(t *testing.T) {
...
+		0x84,             // day:    4
+		0x88,             // hour:   8 utc (18 local)
+		0x8F,             // minute: 15
+		0xAB,             // second: 43
+		0xC6,             // exp:    6 precision units
+		0x0D, 0x2D, 0x06, // nsec:   863494
 	}
 
-	nowish, _ := time.Parse(time.RFC3339Nano, "2019-08-04T18:15:43.863494+10:00")
+	nowish, _ := NewTimestampFromStr("2019-08-04T18:15:43.863494+10:00", Nanosecond, Local)
```
zslayton, line 168: Amusingly, `nowish` is nearly a year old. :)

ion/bits.go
```diff
@@ -244,41 +243,89 @@ func appendTag(b []byte, code byte, len uint64) []byte {
...
+		// Almost certainly two but let's be safe.
+		ret += varUintLen(uint64(utc.dateTime.Year()))
+
+		// Month, day, hour, minute, and second are all guaranteed to be one byte.
+		if utc.precision >= Month {
+			ret++
 
-	// Month, day, hour, minute, and second are all guaranteed to be one byte.
-	ret += 5
+			if utc.precision >= Day {
```
zslayton, line 264: I think this deeply nested `if` tree would be simpler/faster as a `switch` statement.

ion/bitstream.go
```diff
@@ -515,58 +514,86 @@ func (b *bitstream) ReadDecimal() (*Decimal, error) {
...
+		length -= vlen
 		ts[i] = int(val)
 
-		// When i is 3, it means we are setting hour component. A timestamp with
-		// hour, must follow by minute. Hence, len cannot be zero at this point.
-		if i == 3 && len == 0 {
-			return time.Time{},
-				&SyntaxError{"Invalid timestamp - Hour cannot be present without minute", b.pos}
+		// When i is 3, it means we are setting hour component. A timestamp with hour
+		// must follow by minute. Hence, length cannot be zero at this point.
```
zslayton, line 541:
```suggestion
		// When i is 3, it means we are setting the hour component. A timestamp with an hour
		// component must also have a minute component. Hence, length cannot be zero at this point.
```

ion/textwriter_test.go
```diff
@@ -239,8 +239,10 @@ func TestWriteTextDecimal(t *testing.T) {
 func TestWriteTextTimestamp(t *testing.T) {
 	expected := "1970-01-01T00:00:00.001Z\n1970-01-01T01:23:00+01:23"
 	testTextWriter(t, expected, func(w Writer) {
-		w.WriteTimestamp(time.Unix(0, 1000000).In(time.UTC))
-		w.WriteTimestamp(time.Unix(0, 0).In(time.FixedZone("wtf", 4980)))
+		dateTime := time.Unix(0, 1000000).In(time.UTC)
+		w.WriteTimestamp(NewTimestampWithFractionalSeconds(dateTime, Nanosecond, UTC, 3))
+		dateTime = time.Unix(0, 0).In(time.FixedZone("wtf", 4980))
```
zslayton, line 244:
```suggestion
		dateTime = time.Unix(0, 0).In(time.FixedZone("foo", 4980))
```

ion/timestamp.go
```diff
@@ -0,0 +1,303 @@
...
+
+	return ts.dateTime.Equal(ts1.dateTime) &&
+		offset == offset1 &&
+		ts.precision == ts1.precision &&
+		ts.kind == ts1.kind &&
+		ts.numFractionalSeconds == ts1.numFractionalSeconds
+}
+
+// SetLocation sets the location for the internal time object.
+func (ts *Timestamp) SetLocation(loc *time.Location) {
```
fernomac, line 291: I quite like that `time.Time` values are immutable. Would it make sense to follow that for `Timestamp` as well, and change this to `func (ts Timestamp) In(loc *time.Location) Timestamp`?

cmd/ion-go/main.go
justing-bq: Addressed.

ion/bits.go
justing-bq: Addressed.

ion/marshal.go
```diff
@@ -308,8 +307,8 @@ func (m *Encoder) encodeStruct(v reflect.Value) error {
 	}
 
 	t := v.Type()
-	if t == timeType {
-		return m.encodeTime(v)
+	if t == timestampType {
```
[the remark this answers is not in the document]
justing-bq: Looks like this would require us to formally add time.TIme as an ion type. 

@therapon, @zslayton What do you think?

ion/timestamp.go
```diff
@@ -0,0 +1,303 @@
...
+		formatStr := "2006-01-02T15:04:05"
+
+		if precisionUnits > 9 {
+			precisionUnits = 9
+		}
+
+		if precisionUnits > 0 {
+			formatStr += "."
+			for i := uint8(0); i < precisionUnits; i++ {
+				formatStr += "9"
```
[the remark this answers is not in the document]
justing-bq: Done.

ion/timestamp.go
```diff
@@ -0,0 +1,303 @@
...
+
+	dateTime, err := time.Parse(precision.Layout(kind, fractionUnits), dateStr)
+	if err != nil {
+		return Timestamp{time.Time{}, NoPrecision, Unspecified, 0}, err
+	}
+
+	return NewTimestampWithFractionalSeconds(dateTime, precision, kind, fractionUnits), nil
+}
+
+func emptyTimestamp() Timestamp {
```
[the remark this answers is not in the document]
justing-bq: Done.  Thanks for the suggestion.

ion/timestamp.go
```diff
@@ -0,0 +1,303 @@
...
+func NewTimestampWithFractionalSeconds(dateTime time.Time, precision TimestampPrecision, kind TimezoneKind, fractionPrecision uint8) Timestamp {
+	if fractionPrecision > 9 {
+		// 9 is the max precision supported
+		fractionPrecision = 9
+	}
+	return Timestamp{dateTime, precision, kind, fractionPrecision}
+}
+
+// NewTimestampFromStr constructor
+func NewTimestampFromStr(dateStr string, precision TimestampPrecision, kind TimezoneKind) (Timestamp, error) {
```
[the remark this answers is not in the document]
justing-bq: We have `parseTimestamp()` in `textutils.go` which figures out the precision and kind before calling `NewTimestampFromStr()`.

`NewTimestampFromStr()` also allows for precision and kind to differ from what we would infer from the passed in `dateStr`. For example, someone might call `NewTimestampFromStr("2000-05-08T", TimestampPrecisionMinute, TimezoneUnspecified)` to get a Timestamp equivalent to `"2000-05-08T00:00-00:00"`.

ion/timestamp.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	date = date.In(time.FixedZone("fixed", int(offset)*60))
+
+	// Non-zero offset is Local
+	return NewTimestampWithFractionalSeconds(date, precision, Local, fractionPrecision), nil
+}
+
+// Format returns a formatted Timestamp string.
+func (ts *Timestamp) Format() string {
```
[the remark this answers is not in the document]
justing-bq: Done.

ion/timestamp.go
justing-bq: I changed the TImestamp functions to use value receivers. 
Out of curiosity, what is your rationale for preferring immutable value receivers over pointer receivers?

ion/timestamp.go
```diff
@@ -0,0 +1,303 @@
...
+	// Local is for timestamps that have a non-zero offset from UTC (ie. 2001-02-03T04:05+08:30, 2009-05-18T16:20-04:00).
+	Local
+)
+
+// Timestamp struct
+type Timestamp struct {
+	dateTime             time.Time
+	precision            TimestampPrecision
+	kind                 TimezoneKind
+	numFractionalSeconds uint8
```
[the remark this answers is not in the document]
justing-bq: Added getters.

ion/marshal.go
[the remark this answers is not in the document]
zslayton: I'm in favor of also supporting the convenient native types and documenting why those types could be lossy. However, I don't have a strong sense of the level of effort involved. Could we easily add that support in the near future, or does it need to be added now?

> Looks like this would require us to formally add time.TIme as an ion type.

Can you elaborate on this? I read "formally add ... an Ion type" as changing the Ion spec, which is probably not what you're describing.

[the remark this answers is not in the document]
justing-bq: Well implementing this for go time would involve treating time the same way we're treating other ion types such adding a TimeType case to:

image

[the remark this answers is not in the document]
fernomac: I don't think so? The Ion data will still have TimestampType and get routed into decodeTimestampTo; that's already checking the type of v and we can have cases there for both `v.Type() == timeType` and `v.Type() == timestampType`.

ion/timestamp.go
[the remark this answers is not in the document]
fernomac: ParseTimestamp seems like the more common case, so I'd love to have a public version of it to save me some typing. :) I'm fine also having a variant that allows overriding the precision and timezone for folks who want that.

ion/timestamp.go
fernomac: In general, I prefer immutable values for things like this because they're safe to use as constants, reference from different goroutines without synchronization, etc. That's sketchier if I have to worry about someone else changing the timezone out from underneath me. Go follows this pattern a reasonable amount, e.g. with strings and time.Time.

As far as value-type receiver vs pointer-type, it doesn't make a huge difference in practice, but value-type receivers are a signal that the value is immutable and meant to be passed around by value, ala `time.Time`. (This is a pattern I didn't follow with `Decimal`, although in retrospect I arguably should have. I started off basing that one on `*big.Int`, which allows mutation so you can reuse the underlying storage instead of making copies of large byte arrays. I then switched to an immutable style because all the decimals I was working with were small-number-of-digit amounts of money where the copying wasn't an issue, but didn't switch the receivers. :))

ion/timestamp.go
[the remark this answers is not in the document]
justing-bq: I moved the full parse timestamp logic over to `ParseTimestampFromStr()` in `timestamp.go`.

ion/marshal.go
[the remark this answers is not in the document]
justing-bq: Supporting this will require more effort than I want to include in this PR so I created a new issue for it.

https://github.com/amzn/ion-go/issues/113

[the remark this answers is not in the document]
zslayton: Thanks for opening an issue, I think we can add that in another PR.

ion/timestamp.go
justing-bq: Makes sense. Thanks for the explanation.

PR #99, opened 2020-07-22
Preserve annotations when unmarshalling values

Armanbqt:
```Decoder``` drops annotations and consequently the output of Unmarshal does not have any of the annotations.

To get the annotations when Unmarshalling, user should send a struct to ```unmarshal()``` whereby one field of the struct is tagged as ``` `ion:"annotations"` ```.

For instance:
```	
type foo struct {
   Value   int
   AnyName []string `ion:"annotations"`
}
...
var val foo
err := UnmarshalStr("age::2, &val)
```

### CODE REVIEW

File: ion/unmarshal.go [hunk 1]
```diff
@@ -671,3 +701,48 @@ func indirect(v reflect.Value, wantPtr bool) reflect.Value {
...
 }
+
+func (d *Decoder) decodeToStructWithAnnotation(v reflect.Value) error {
+	if err := d.attachAnnotations(v); err != nil {
+		return err
+	}
+
+	fields := fieldsFor(v.Type())
+	for _, field := range fields {
+		if field.name != AnnotationsTag {
```
fernomac, line 712: Trying to stash the value into the first field that isn't named "annotations" feels like it could be error-prone. Would it make sense to require the value-receiving field to be similarly annotated with something like `ion:",value"`?

ion/unmarshal.go
```diff
@@ -671,3 +701,48 @@ func indirect(v reflect.Value, wantPtr bool) reflect.Value {
...
+				break
+			}
+		}
+	}
+	return nil
+}
+
+func (d *Decoder) attachAnnotations(v reflect.Value) error {
+	fields := fieldsFor(v.Type())
+	field := findField(fields, AnnotationsTag)
```
fernomac, line 732: I think we want to look for a field with an "annotations" opt (in the style of `omitempty`), rather than relying on the field's name? A struct might legitimately have a field named "annotations" which would lead to surprising behavior.

```go
type LegalDocument struct {
  Text string
  Annotations string  // oops this will explode when we try to set it to a []string
}
```

ion/unmarshal.go
```diff
@@ -671,3 +701,48 @@ func indirect(v reflect.Value, wantPtr bool) reflect.Value {
...
+			return err
+		}
+		annotations := d.r.Annotations()
+		annotationsType := indirect(reflect.ValueOf(annotations), false).Type()
+		valueSlice := reflect.MakeSlice(annotationsType, 0, len(annotations))
+
+		for _, an := range annotations {
+			valueSlice = reflect.Append(valueSlice, reflect.ValueOf(an))
+		}
+		subValue.Set(valueSlice)
```
fernomac, line 745: Any reason we can't just do `subValue.Set(reflect.ValueOf(annotations))` here?

ion/unmarshal_test.go
```diff
@@ -568,3 +568,87 @@ func TestDecodeLotsOfInts(t *testing.T) {
...
+				t.Errorf("expected %v, got %v", eval, val)
+			}
+		})
+	}
+
+	test("with::multiple::annotations::[1, 2, 3]", "list", foo{[]int{1, 2, 3}, []string{"with", "multiple", "annotations"}})
+	test("with::multiple::annotations::(1 2 3)", "sexp", foo{[]int{1, 2, 3}, []string{"with", "multiple", "annotations"}})
+}
+
+func TestUnmarshalNestedStructsWithAnnotation(t *testing.T) {
```
fernomac, line 628: Is there a legitimate use-case for this? Seems to me like I'd only do this if I was expecting the annotated value to be an actual struct, and I'd explicitly want a decoding error if I got that wrong?

ion/unmarshal.go
```diff
@@ -250,6 +258,9 @@ func (d *Decoder) decodeBoolTo(v reflect.Value) error {
 		v.SetBool(val)
 		return nil
 
+	case reflect.Struct:
+		return d.decodeToStructWithAnnotation(v)
```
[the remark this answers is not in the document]
Armanbqt: This check is deferred to later when ```d.decodeTo()``` in ```decodeToStructWithAnnotation()``` tries to decode (in this case) a bool value in a non-boolean field.

ion/unmarshal.go
```diff
@@ -307,8 +318,9 @@ func (d *Decoder) decodeIntTo(v reflect.Value) error {
 				return err
 			}
 			v.Set(reflect.ValueOf(*val))
-			return nil
+			return d.attachAnnotations(v)
```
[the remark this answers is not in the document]
Armanbqt: You are right. Good catch, thanks!
I'll remove it and and add a test for it.

ion/unmarshal.go
Armanbqt: Agree, to some extent. Randomly grabbing the first field (except "annotation") to put the data in, for sure, is what I do not like about about my solution, but apart from not having a better idea:
- I did not want to impose another restriction on user, to tag the field for value as well as the one for annotations.
- I cannot think of a case where a struct need to have more than 2 fields (1 for annotations and the other for the value)

ion/unmarshal.go
Armanbqt: That is right; however, I am not sure if I understood your suggestion correctly.
Are you suggesting to change ```annotations``` to ```,annotations```? Or adding another field to the expected struct?

ion/unmarshal.go
Armanbqt: The only reason was my lack of familiarity with Go!
Modified, thank.

ion/unmarshal_test.go
Armanbqt: I know a user of Ion-Go is testing some documents with nested structs and I wanted to have a test for that as well.
Yet, I do not know if there is a legitimate use-case.

ion/unmarshal.go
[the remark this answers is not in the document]
fernomac: And if the struct happens to have a bool field? It feels weird to silently map a bool Ion value into the struct unless the user has explicitly marked that that's the behavior they intend in some way. (And, in particular, that they'll get matching behavior if they encode the value, once that half of it is implemented.)

ion/unmarshal.go
fernomac: Maybe leave the value field untagged but verify before taking this codepath that the struct has exactly two fields, one of type `[]string` tagged with `ion:"annotations"` and the other of which will receive the value?

ion/unmarshal.go
fernomac: Correct, users would do `ion:",annotations"` (the actual field name doesn't matter since the struct's not actually getting serialized). The `field` struct would get an extra bool field named `annotations` that would get filled in like `omitempty` does; this code would look for a field with that opt set to true.

ion/unmarshal_test.go
fernomac: Nested structs in general, definitely. Unmarshaling the value `2` into a series of nested structs seems pretty strange though. Was the `str` here supposed to be `"with::multiple::annotations::{Value:2}"` and the test case was passing by accident? :)

ion/unmarshal.go
```diff
@@ -318,7 +318,6 @@ func (d *Decoder) decodeIntTo(v reflect.Value) error {
 				return err
 			}
 			v.Set(reflect.ValueOf(*val))
-			return d.attachAnnotations(v)
```
fernomac: I think you want to go back to `return nil`?

ion/unmarshal.go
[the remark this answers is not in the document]
Armanbqt: Updated the PR.
Thanks.

Adding annotations in Marshal is being implemented too.

ion/unmarshal.go
Armanbqt: Updated the PR.
Thanks.

ion/unmarshal.go
Armanbqt: Updated the PR.
Thanks.

Armanbqt:
Created #103 for nested structs.

ion/unmarshal_test.go
Armanbqt: Updated the implementation and the test ([this commit](9cd2cc5f1bf59e0708f2bacd635988747a2c8341) specifically)

ion/unmarshal_test.go
```diff
@@ -568,3 +568,105 @@ func TestDecodeLotsOfInts(t *testing.T) {
...
+	type innerStruct struct {
+		Value           int
+		ValueAnnotation []string `ion:",annotations"`
+	}
+
+	type mainStruct struct {
```
zslayton, line 636: To me, `mainStruct` and `topLevelStruct` mean the same thing, and `innerStruct` isn't actually an Ion struct.

How about `topLevelStruct`, `nestedStruct`, and `nestedInt`?

ion/unmarshal.go
```diff
@@ -671,3 +698,86 @@ func indirect(v reflect.Value, wantPtr bool) reflect.Value {
...
+			if isAcceptableKind(listofkinds, subValue.Type().Kind()) {
+				acceptableValueType = true
+			}
+		}
+	}
+	return hasAnnotation && acceptableValueType, nil
+}
+
+func isAcceptableKind(valueAcceptableKinds []reflect.Kind, valueKind reflect.Kind) bool {
+	for _, kind := range valueAcceptableKinds {
```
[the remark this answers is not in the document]
Armanbqt: Opened #110

ion/unmarshal_test.go
Armanbqt: Updated.

PR #104, opened 2020-07-25
Add annotations when Marshalling values

Armanbqt:
Encoder does not take annotations and consequently Marshal cannot serialize values with annotations.

To get the annotations when Marshalling, user should send a struct to Marshal() whereby one field of the struct is a slice of string and tagged as `ion:",annotations"`.

For instance:
```
type foo struct {
   Value   int
   AnyName []string `ion:",annotations"`
}
...
val := foo{5, []string{"age"}}
buf := MarshalText(val)
```

### DIFF
--- a/ion/marshal.go
+++ b/ion/marshal.go
@@ -369,7 +369,8 @@ func (m *Encoder) encodeWithAnnotation(v reflect.Value) error {
 				return err
 			}
 			if annotations.Kind() != reflect.Slice {
-				return fmt.Errorf("ion: '%v' is provided for annotations, it must be []string", annotations.Kind())
+				return fmt.Errorf("ion: '%v' is provided for annotations,"+
+					"it must be of type []string", annotations.Kind())
 			}
 			listOfAnnotations := annotations.Interface().([]string)
 			err = m.w.Annotations(listOfAnnotations...)
--- a/ion/marshal.go
+++ b/ion/marshal.go
@@ -179,7 +179,7 @@ func (m *Encoder) encodeValue(v reflect.Value, hint Type) error {
 		return m.encodePtr(v, hint)
 
 	case reflect.Struct:
-		return m.encodeStruct(v, hint)
+		return m.encodeStruct(v)
 
 	case reflect.Map:
 		return m.encodeMap(v, hint)
@@ -299,7 +299,14 @@ func (m *Encoder) encodeArray(v reflect.Value, hint Type) error {
 }
 
 // EncodeStruct encodes a struct to the output writer as an Ion struct.
-func (m *Encoder) encodeStruct(v reflect.Value, hint Type) error {
+func (m *Encoder) encodeStruct(v reflect.Value) error {
+	fields := fieldsFor(v.Type())
+	for _, field := range fields {
+		if field.annotations {
+			return m.encodeWithAnnotation(v)
+		}
+	}
+
 	t := v.Type()
 	if t == timeType {
 		return m.encodeTime(v)
@@ -308,9 +315,9 @@ func (m *Encoder) encodeStruct(v reflect.Value, hint Type) error {
 		return m.encodeDecimal(v)
 	}
 
-	fields := fieldsFor(v.Type())
-
-	m.w.BeginStruct()
+	if err := m.w.BeginStruct(); err != nil {
+		return err
+	}
 
 FieldLoop:
 	for i := range fields {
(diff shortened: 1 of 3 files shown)


ion/marshal.go
```diff
@@ -352,6 +359,30 @@ func (m *Encoder) encodeDecimal(v reflect.Value) error {
...
+	original := v
+	fields := fieldsFor(v.Type())
+	for _, field := range fields {
+		if field.annotations {
+			annotations, err := findSubvalue(original, &field)
+			if err != nil {
+				return err
+			}
+			if annotations.Kind() != reflect.Slice {
+				return fmt.Errorf("ion: '%v' is provided for annotations, it must be []string", annotations.Kind())
```
therapon, line 372:
```suggestion
				return fmt.Errorf("ion: '%v' is provided for annotations, it must be of type []string", annotations.Kind())
```

ion/marshal_test.go
```diff
@@ -285,3 +285,38 @@ func TestMarshalCustomMarshaler(t *testing.T) {
...
+	test(buildValue(true), "bool", "multiple::annotations::true")
+	test(buildValue(5), "int", "multiple::annotations::5")
+	test(buildValue(float32(math.MaxFloat32)), "float", "multiple::annotations::3.4028234663852886e+38")
+	test(buildValue(MustParseDecimal("1.2")), "decimal", "multiple::annotations::1.2")
+	test(buildValue(time.Date(2000, 1, 2, 3, 4, 5, 0, time.UTC)),
+		"timestamp", "multiple::annotations::2000-01-02T03:04:05Z")
+	test(buildValue("stringValue"), "string", "multiple::annotations::\"stringValue\"")
+	test(buildValue([]byte{4, 2}), "blob", "multiple::annotations::{{BAI=}}")
+	test(buildValue([]int{3, 5, 7}), "list", "multiple::annotations::[3,5,7]")
+	test(buildValue(map[string]int{"b": 2, "a": 1}), "struct", "multiple::annotations::{a:1,b:2}")
```
therapon, line 321: Could we add a test where the annotation is a symbol that contains a space, e.g., `'i am an annotation'::also::true`

Armanbqt: Changed the first annotation to a symbol for all the examples.

ion/marshal.go
Armanbqt: Oops, I did not see the suggested commit, I included the change in [this commit](039acd2f9ab1ad204e61ef1f78a4c2186b377d49).

ion/marshal.go
```diff
@@ -352,6 +359,31 @@ func (m *Encoder) encodeDecimal(v reflect.Value) error {
 	return m.w.WriteDecimal(d)
 }
 
+func (m *Encoder) encodeWithAnnotation(v reflect.Value) error {
+	original := v
+	fields := fieldsFor(v.Type())
```
fernomac, line 364: Might be worth passing these in instead of re-deriving them here?

ion/marshal.go
```diff
@@ -352,6 +359,31 @@ func (m *Encoder) encodeDecimal(v reflect.Value) error {
...
+	for _, field := range fields {
+		if field.annotations {
+			annotations, err := findSubvalue(original, &field)
+			if err != nil {
+				return err
+			}
+			if annotations.Kind() != reflect.Slice {
+				return fmt.Errorf("ion: '%v' is provided for annotations,"+
+					"it must be of type []string", annotations.Kind())
```
fernomac, line 373: nit: space after comma at the end of line 372?

ion/marshal.go
```diff
@@ -352,6 +359,31 @@ func (m *Encoder) encodeDecimal(v reflect.Value) error {
...
+			}
+			listOfAnnotations := annotations.Interface().([]string)
```
fernomac, line 375: You can combine this and a more-accurate version of the above typecheck by using the [two-return-value type assertion](https://tour.golang.org/methods/15):

```go
annos, ok := annotations.Interface().([]string)
if !ok {
  return fmt.Errorf(...)
}
```

Armanbqt: Thanks for suggesting it. Updated.

ion/marshal.go
Armanbqt: Done.

ion/marshal.go
Armanbqt: Done.

PR #105, opened 2020-07-27
Add tests to Unmarshal for Ion binary

Armanbqt:
Expand unit testing on Unmarshal - Ion binary test data

### DIFF
--- a/ion/unmarshal_test.go
+++ b/ion/unmarshal_test.go
@@ -28,6 +28,7 @@ func TestUnmarshalBool(t *testing.T) {
 	test("true", true)
 	test("false", false)
 }
+
 func TestUnmarshalBoolPtr(t *testing.T) {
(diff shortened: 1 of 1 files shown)


fernomac:
Do we expect these tests to hit new codepaths in the unmarshal code that we can't hit with a text reader? My assumption (hope?) was that the marshal/unmarshal code would be agnostic of the underlying reader/writer implementation (whether text, binary, or third-party implementation of the interfaces) and so the text reader is effectively a convenient way to mock the reader interface.

Armanbqt:
Hi @fernomac , these tests are not hitting new codepaths, however, when we were adding binary tests to Marshal, it let us to iron out some issues we have not found before, such as #101, #107 and #108.
These tests are to give us more confident on that regard.

PR #124, opened 2020-07-31
Improve documentation on Marshal and Unmarshal

Armanbqt:
This is not for Preview
----

This pull request is to improve documentation on Marshal and Unmarshal

### DIFF
--- a/ion/unmarshal.go
+++ b/ion/unmarshal.go
@@ -33,8 +33,7 @@ var (
 // Unmarshal unmarshals Ion data to the given object.
 //
 // User must pass the proper object type to the unmarshalled Ion data.
-// unmarshalled to is important. Below is the mapping between Go native
-// type and Ion types. e.g.,
+// Below is the mapping between Go native type and Ion types. e.g.,
 //
 //     boolBytes := []byte{0xE0, 0x01, 0x00, 0xEA, 0x11}
 //     var boolVal bool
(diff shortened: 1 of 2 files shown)


ion/unmarshal.go
```diff
@@ -31,6 +31,60 @@ var (
...
+//		ion.Decimal/interface{}							decimal
+//		ion.Timestamp/interface{}						timestamp
+//		string/interface{}								symbol
+//		string/interface{}								string
+//		[]byte/[]interface{}{}							clob
+//		[]byte/[]interface{}{}							blob
+//		[]interface{}{}									list
+//		[]interface{}{}									sexp
+//		map[string]interface{}{}/struct/interface{}		struct
+//
```
Armanbqt, line 87: I am not sure why this is shown like this here. In my IDE is looks well aligned and I was careful about the spacing:

unmarshal

ion/unmarshal.go
```diff
@@ -31,6 +31,60 @@ var (
 )
 
 // Unmarshal unmarshals Ion data to the given object.
+//
+// User must pass the proper object type to the unmarshalled Ion data.
+// unmarshalled to is important. Below is the mapping between Go native
```
therapon, line 36: this looks like a sentence fragment

ion/unmarshal.go
therapon: My guess is that the whitespace has tabs and spaces. Stick to one, I would recommend spaces.

PR #14, opened 2020-02-23
Adds SymbolToken and SymbolTable APIs.

almann:
Removes non-used `symbols.go` module which had a preliminary symbol
table implementation.

Adds SymbolToken as a pure data structure.

Adds SymbolTable as an abstract data type for
system/shared/local symbol tables.

None of the APIs are currently exported for making symbol tables.

Resolves #11

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #127, opened 2020-08-07
Design and stub Symbol Token API

byronlin-bq:
Issue #, if available:

Description of changes:


ion/symboltoken.go
```diff
@@ -0,0 +1,46 @@
...
+func (st *SymbolToken) Sid() int64 {
+	return st.sid
+}
+
+// Gets the text of this symbol token.
+func (st *SymbolToken) Text() *string {
+	return st.text
+}
+
+func (st *SymbolToken) Equal(o *SymbolToken) bool {
```
Armanbqt, line 39: Same as ```importLocation.Equal```

ion/symboltoken.go
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package ion
+
+var UnknownSid int64 = -1
```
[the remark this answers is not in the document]
fernomac: Even better: `const UnknownSID = -1`. :)

(Actually, could you use 0 here instead of -1? I think so, and that'd play nicely in that `SymbolToken{}` would be the right way to initialize a default, invalid symbol token.)

ion/importlocation.go
```diff
@@ -0,0 +1,32 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package ion
+
+// TODO comment
+type ImportLocation struct {
+	importName *string
```
fernomac, line 20: As with my feedback on #93, pointers to strings feel weird and unidiomatic in go. Most notably, it means you can't call NewImportLocation as a one-liner with an inline name, you've got to allocate a new string object on the heap. Would it be possible to have this just be a normal `string`, or will importName ever be meaningfully nil?

You also probably want to either add accessor methods for importName and sid, or (my vote) just export them and get rid of `NewImportLocation`.

ion/symboltoken.go
```diff
@@ -0,0 +1,46 @@
...
+
+// A SymbolToken providing both the symbol text and the assigned symbol ID.
+// Symbol tokens may be interned into a SymbolTable.
+// A text=nil or sid=-1 value might indicate that such field is unknown in the contextual symbol table.
+type SymbolToken struct {
+	text           *string
```
fernomac, line 24: Ditto on `*string`[1] and exporting these if `NewSymbolToken` and the accessors aren't going to do anything but set/get them.

[1] Although here I suppose you want to be able to distinguish between a symbol whose text is known to be the empty string and a symbol whose text is not known because of a missing shared symbol table? Ugh, I once again wish we hadn't made the empty string a valid symbol. :/

ion/symboltoken.go
```diff
@@ -0,0 +1,46 @@
...
+	sid            int64
+	importLocation *ImportLocation
+}
+
+// Gets the ID of this symbol token.
+func (st *SymbolToken) Sid() int64 {
```
fernomac, line 30: Nitpick: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

byronlin-bq:
> Even better: const UnknownSID = -1. :)

(Actually, could you use 0 here instead of -1? I think so, and that'd play nicely in that SymbolToken{} would be the right way to initialize a default, invalid symbol token.)

Updated to const.

We have to use unknown SID as -1 because 0 is for symbol zero which is used to denote that a SymbolToken has unknown text in any symbol table.

ion/symboltoken.go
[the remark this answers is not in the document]
byronlin-bq: Updated to const.

We have to use unknown SID as -1 because 0 is for symbol zero which is used to denote that a SymbolToken has unknown text in any symbol table.

ion/importlocation.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+// NewImportLocation creates an ImportLocation struct.
+func NewImportLocation(importName *string, sid int64) *ImportLocation {
+	return &ImportLocation{importName, sid}
+}
+
+// TODO comment
+func (il *ImportLocation) Equal(o *ImportLocation) bool {
+	return *il.importName == *o.importName && il.sid == o.sid
```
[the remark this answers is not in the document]
byronlin-bq: Updated to check for nil and added unit tests

ion/symboltoken.go
byronlin-bq: Updated to check for nil and added unit tests

ion/symboltoken.go
byronlin-bq: Updated to SID

ion/symboltoken.go
byronlin-bq: > Ditto on *string[1] and exporting these if NewSymbolToken and the accessors aren't going to do anything but set/get them.

Updated to export the struct attributes and removed the setters/getters.

> [1] Although here I suppose you want to be able to distinguish between a symbol whose text is known to be the empty string and a symbol whose text is not known because of a missing shared symbol table? Ugh, I once again wish we hadn't made the empty string a valid symbol. :/

Thats correct, we need nil string to represent as unknown text.

ion/importlocation.go
byronlin-bq: > As with my feedback on #93, pointers to strings feel weird and unidiomatic in go. Most notably, it means you can't call NewImportLocation as a one-liner with an inline name, you've got to allocate a new string object on the heap. Would it be possible to have this just be a normal string, or will importName ever be meaningfully nil?

When SymbolToken is created with default ImportLocation, it will look like: 
```
ImportLocation {
  ImportName: nil
  SID: 0
}
```

This is used to represent an undefined ImportLocation. 



> You also probably want to either add accessor methods for importName and sid, or (my vote) just export them and get rid of NewImportLocation.

Removed `NewImportLocation` and exported the struct attributes

byronlin-bq: Thinking about this more, we can set SymbolToken ImportLocation to be a pointer and the default ImportLocation will be nil. Setting ImportLocation to a normal string

byronlin-bq:
Updated this PR to use a `[symboltoken API PR](https://github.com/amzn/ion-go/pull/14/files?file-filters%5B%5D=.go&hide-deleted-files=true)` that was merged before this repo was seeded with fernomac's [ion-go repo ](https://github.com/fernomac/ion-go)

ion/symboltoken.go
```diff
@@ -0,0 +1,89 @@
...
+const (
+	// SymbolIDUnknown is placeholder for when a symbol token has no symbol ID.
+	SymbolIDUnknown = -1
+)
+
+// ImportSource is a reference to a SID within a shared symbol table.
+type ImportSource struct {
+	// The name of the shared symbol table that this token refers to.
+	Table string
+	// The ID of the interned symbol text within the shared SymbolTable.
```
justing-bq, line 29: I think it would be more readable with an empty line in between struct components when we are commenting on each of them.

ion/symboltoken.go
```diff
@@ -0,0 +1,89 @@
...
+// Equal figures out if two import sources are equal for each component.
+func (is *ImportSource) Equal(o *ImportSource) bool {
+	return is.Table == o.Table && is.SID == o.SID
+}
+
+// SymbolToken are the values that annotations, field names, and the textual content of Ion symbol values.
+// The `nil` value for SymbolToken is $0.
+type SymbolToken struct {
+	// The string text of the token or nil if unknown.
+	Text *string
```
justing-bq, line 51: Likewise here.

ion/symboltoken.go
byronlin-bq: Updated

ion/symboltoken.go
byronlin-bq: Updated

ion/symboltoken.go
```diff
@@ -0,0 +1,91 @@
...
+	source := "nil"
+	if st.Source != nil {
+		source = fmt.Sprintf("{%q %d}", st.Source.Table, st.Source.SID)
+	}
+
+	return fmt.Sprintf("{%s %d %s}", text, st.LocalSID, source)
+}
+
+// Equal figures out if two symbol tokens are equal for each component.
+func (st *SymbolToken) Equal(o *SymbolToken) bool {
```
therapon, line 83: What is the intent of this `Equals`? The code checks 2 of the 3 fields in `SymbolToken`.

ion/symboltoken.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+// ImportSource is a reference to a SID within a shared symbol table.
+type ImportSource struct {
+
+	// The name of the shared symbol table that this token refers to.
+	Table string
+
+	// The ID of the interned symbol text within the shared SymbolTable.
+	// This must be greater than 1.
```
justing-bq, line 32: Do you mean "greater or equal to 1"?

ion/symboltoken.go
```diff
@@ -0,0 +1,91 @@
...
+		source = fmt.Sprintf("{%q %d}", st.Source.Table, st.Source.SID)
+	}
+
+	return fmt.Sprintf("{%s %d %s}", text, st.LocalSID, source)
+}
+
+// Equal figures out if two symbol tokens are equal for each component.
+func (st *SymbolToken) Equal(o *SymbolToken) bool {
+	if st.Text == nil || o.Text == nil {
+		if st.Text == nil && o.Text == nil && st.LocalSID == o.LocalSID {
```
justing-bq, line 85: Lets simplify this to `return st.Text == nil && o.Text == nil && st.LocalSID == o.LocalSID`

justing-bq: Actually this entire method can be simplified to 
`return ((st.Text == nil && o.Text == nil) || (st.Text != nil && o.Text != nil && *st.Text == *o.Text)) && st.LocalSID == o.LocalSID`

justing-bq:
Can we add a constructor for symbol token?

ion/symboltoken.go
byronlin-bq: Updated `Equal` and unit tests to match symbol token equivalence spec in: http://amzn.github.io/ion-docs/guides/symbols-guide.html

ion/symboltoken.go
byronlin-bq: Updated `Equal` function to match symbol token equivalency spec in http://amzn.github.io/ion-docs/guides/symbols-guide.html

ion/symboltoken.go
byronlin-bq: Updated

byronlin-bq:
> Can we add a constructor for symbol token?

I originally had a constructor for symbol token but removed it because we have any special logic we when are constructing the symbol token

fernomac:
> > Can we add a constructor for symbol token?
> 
> I originally had a constructor for symbol token but removed it because we have any special logic we when are constructing the symbol token

Generally speaking yeah, it's more go-ey to just use the inline `&SymbolToken{...}` syntax. (Ditto for `newSource` now that `ImportSource` has exported fields.) In this case, if you expect users to create SymbolTokens (I don't think so but who knows?), it might be nice to have a convenience func to work around the use of `*string`--take a `string` parameter and handle the deref for the common case where the text is known.

justing-bq:
Could we get symbol token constructors similar to 

https://github.com/amzn/ion-java/blob/master/src/com/amazon/ion/impl/_Private_Utils.java#L208
and
https://github.com/amzn/ion-java/blob/master/src/com/amazon/ion/impl/_Private_Utils.java#L240 ?

byronlin-bq:
> Could we get symbol token constructors similar to
> 
> https://github.com/amzn/ion-java/blob/master/src/com/amazon/ion/impl/_Private_Utils.java#L208
> and
> https://github.com/amzn/ion-java/blob/master/src/com/amazon/ion/impl/_Private_Utils.java#L240 ?

Added in new #145 PR.

Closing this PR instead

PR #128, opened 2020-08-10
Fix S-Expression Tests

justing-bq:

### DIFF
--- a/ion/textreader.go
+++ b/ion/textreader.go
@@ -213,7 +213,7 @@ func (t *textReader) nextBeforeTypeAnnotations() (bool, error) {
 		}
 		return false, &UnexpectedEOFError{t.tok.Pos() - 1}
 
-	case tokenSymbolOperator, tokenDot, tokenDotSymbol:
+	case tokenSymbolOperator, tokenDot:
 		if t.ctx.peek() != ctxInSexp {
 			// Operators can only appear inside an sexp.
 			return false, &UnexpectedTokenError{tok.String(), t.tok.Pos() - 1}
--- a/ion/textreader.go
+++ b/ion/textreader.go
@@ -213,7 +213,7 @@ func (t *textReader) nextBeforeTypeAnnotations() (bool, error) {
 		}
 		return false, &UnexpectedEOFError{t.tok.Pos() - 1}
 
-	case tokenSymbolOperator, tokenDot:
+	case tokenSymbolOperator, tokenDot, tokenDotSymbol:
 		if t.ctx.peek() != ctxInSexp {
 			// Operators can only appear inside an sexp.
 			return false, &UnexpectedTokenError{tok.String(), t.tok.Pos() - 1}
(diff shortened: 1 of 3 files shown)


ion/textreader.go
```diff
@@ -237,7 +237,11 @@ func (t *textReader) nextBeforeTypeAnnotations() (bool, error) {
 				if err := t.verifyUnquotedSymbol(val, "annotation"); err != nil {
 					return false, err
 				}
+			} else if tok == tokenSymbolOperator {
+				return false, &SyntaxError{
+					"annotations with operator " + val + " should be enclosed in quotes", t.tok.Pos() - 1}
```
zslayton, line 242: Quoting from [the `symbols` section of the spec](https://amzn.github.io/ion-docs/docs/spec.html#symbol):

> Within S-expressions, the rules for unquoted symbols include another set of tokens: operators. An operator is an unquoted sequence of one or more of the following nineteen ASCII characters: !#%&*+-./;<=>?@^`|~ 

AFAIK, the concept of an 'operator' only exists within the context of s-expressions, which makes this error message rather confusing. I would generalize it a bit.

```suggestion
			} else if tok == tokenSymbolOperator {
				return false, &SyntaxError{
					"Annotations that include a '" + val + "' must be enclosed in quotes.", t.tok.Pos() - 1}
```

ion/tokenizer.go
```diff
@@ -355,6 +361,8 @@ func (t *tokenizer) ReadValue(tok token) (string, error) {
 		str, err = t.readQuotedSymbol()
 	case tokenSymbolOperator, tokenDot:
 		str, err = t.readOperator()
+	case tokenDotSymbol:
+		str, err = t.readDotSymbol()
```
zslayton, line 365: I'm somewhat confused by this: a `.` is an operator when it appears in an s-expression. (That is to say: it's a symbol.) Otherwise, it's a plain token.

Shouldn't the `tokenSymbolOperator` and `tokenDot` cases cover this? Are there no other operators that require similar redundancy?

justing-bq: Simplified this to no longer use tokenDotSymbol.

ion/textreader.go
```diff
@@ -237,7 +237,11 @@ func (t *textReader) nextBeforeTypeAnnotations() (bool, error) {
 				if err := t.verifyUnquotedSymbol(val, "annotation"); err != nil {
 					return false, err
 				}
+			} else if tok == tokenSymbolOperator {
+				return false, &SyntaxError{
+					"Annotations that include a '" + val + "' must be enclosed in quotes.", t.tok.Pos() - 1}
```
zslayton, line 242: @fernomac just [informed me on another PR](https://github.com/amzn/ion-go/pull/130#discussion_r469717690) that:

> per https://github.com/golang/go/wiki/CodeReviewComments#error-strings error strings in go should not start with a capitalized word or end in punctuation

Sorry, I know this was my suggested text. 🤦 Rather than fixing this here, I think we should make a cleanup issue to fix this project-wide before a 1.0 release.

zslayton: Opened #131 based on this comment.

Issue #11, opened 2020-01-31
Symbol Token and Symbol Table


almann:
Had a a whiteboard session with @pbcornell and @mrutter-amzn on 1/24/2020 with some consensus on how to model `SymbolToken` and `ImportSource`:

![20200124_113115](https://user-images.githubusercontent.com/503506/73557801-b7101b00-4406-11ea-86db-01270a98d969.jpg)

The variants of local/system/shared symbol table are closed so we should consider implementing this through a `struct` rather than a `interface` type.

One thing that came out of this discussion that we should consider in implementing this is putting this into a `symbols` package and renaming the above APIs accordingly.

tgregg:
Why is 'version' needed in 'ImportSource'?

tgregg:
Since 'LocalSid' is present in 'SymbolToken', it must be used as a caching mechanism within a particular local symbol table context. What is the mechanism for guaranteeing that 'LocalSid' does not leak into a different context?

almann:
> Why is 'version' needed in 'ImportSource'?

You are right, this is an oversight from our whiteboard design.

almann:
> Since 'LocalSid' is present in 'SymbolToken', it must be used as a caching mechanism within a particular local symbol table context. What is the mechanism for guaranteeing that 'LocalSid' does not leak into a different context?

I am a little confused about the question.  I do believe the `SymbolToken` data structure should have the local SID for completeness and optimization when transferring between different streams with compatible contexts.

That said, are you saying this should *not* be modeled or are you asking about a constraint that should exist elsewhere (or enforced by the data structure)?

tgregg:
Let's first agree that LocalSID, from a correctness perspective, is not necessary. That's because
1. If the text is known, the LocalSID can always be calculated (or a new one assigned).
2. If the text is not known, and there is an ImportSource, then the LocalSID can be calculated in any symbol table context that includes a compatible shared symbol table import.
3. If the text is not known and there is no ImportSource, then this represents a symbol with unknown text from a local symbol table, which is always equivalent to symbol zero.

If we agree on that, then we can agree that LocalSID *may* be used to cache the calculations referred to in 1. and 2. above. This is great when the SymbolToken is always used in compatible symbol table contexts because it reduces repetitive recalculation of the local SID in that context.

[...]

PR #129, opened 2020-08-11
Incorporate SymbolToken to SymbolTable

byronlin-bq:
Issue #, if available:

Description of changes:

Add following APIs to SymbolTable.go
```
Find(symbol string)
```

This PR is dependent on #127 


ion/symboltable.go
```diff
@@ -28,7 +28,8 @@ type SymbolTable interface {
 	Symbols() []string
 	// MaxID returns the maximum ID this symbol table defines.
 	MaxID() uint64
-
+	// Find finds the the SymbolToken by its name.
+	Find(symbol string) (SymbolToken, bool)
```
fernomac, line 32: Should this return a `*SymbolToken`?

ion/symboltable.go
```diff
@@ -37,6 +38,8 @@ type SymbolTable interface {
 	WriteTo(w Writer) error
 	// String returns an ion text representation of the symbol table.
 	String() string
+	// InternToken adds a new symbol to this symbol table, or finds an existing definition of it.
+	InternToken(symbol string) SymbolToken
```
fernomac, line 42: `SymbolTable`s are meant to be immutable. Could this go on `SymbolTableBuilder` instead?

byronlin-bq: Removed this API because symboltablebuilder has `Add()`

ion/symboltable.go
byronlin-bq: Updated to return *symbolToken.

ion/symboltable.go
```diff
@@ -128,6 +129,20 @@ func (s *sst) Adjust(maxID uint64) SharedSymbolTable {
...
+	if !ok {
+		return nil
+	}
+
+	text, ok := s.FindByID(id)
+	if !ok {
+		return nil
+	}
+
+	return &SymbolToken{Text: &text, LocalSID: SymbolIDUnknown}
```
justing-bq, line 143: Why are we setting the SID to `SymbolIDUnknown`? What about the `id` that we computed earlier?

ion/symboltable.go
```diff
@@ -318,6 +337,24 @@ func (t *lst) MaxID() uint64 {
...
+		symbolToken := imp.Find(s)
+		if symbolToken != nil {
+			return symbolToken
+		}
+	}
+
+	// Check local
+	_, ok := t.index[s]
+	if ok {
+		return &SymbolToken{Text: &s, LocalSID: SymbolIDUnknown}
```
justing-bq, line 352: Again why `SymbolIDUnknown`

ion/symboltable.go
byronlin-bq: According to the Ion specs, it is irrelevant

> SymbolToken APIs, return a SymbolToken with that text and with an undefined importLocation (it is irrelevant).

http://amzn.github.io/ion-docs/guides/symbols-guide.html

byronlin-bq: According to Almann: https://github.com/amzn/ion-go/issues/11#issuecomment-580951115

> If the Local SID is defined, the text must be unknown and the source must be undefined. This implies that the symbol token has yet to be resolved and we do not (yet) know the context of such a token.
In all other cases Local SID must be undefined as this implies that the symbol token has been resolved and the local identifier is not meaningful in this context.

ion/symboltable.go
byronlin-bq: same as above ^

byronlin-bq:
Moved this PR into #145

PR #130, opened 2020-08-12
Fix malformed clobs

Armanbqt:
This is to fix various cases where a ```clob``` should be considered invalid:
- Check for ```\u``` and ```\U``` in [this commit](402a469).
- Check for non 7-bit ASCII characters in [this commit](d946a09)
- Check for comments inside a clob in [this commit](a4cda0f)


ion/tokenizer.go
```diff
@@ -1379,3 +1383,15 @@ func isNewLineChar(c int) bool {
 	return c == 0x0A || //new line
 		c == 0x0D //carriage return
 }
+
+func isClobWithInvalidChar(clob bool, c int) bool {
+	// The string in clob may only contain 7-bit ASCII characters.
+	return clob && c > 0x7f
```
zslayton, line 1389: [The clob section](https://amzn.github.io/ion-docs/docs/spec.html#clob) of the primary spec page says:

> The string may only contain legal 7-bit ASCII characters, using the same escaping syntax as string and symbol values.

However, there's an important clarification buried on the [Strings and Clobs](https://amzn.github.io/ion-docs/docs/stringclob.html#ion-clob) page:

> Non-printable ASCII and non-ASCII Unicode code points are not allowed un-escaped in the string bodies.

This means that ASCII characters below `31` (`0x1F`) must also be escaped because they are non-printable. The same page provides a listing of alternative escapes (`BEL` / `0x07` / `\a`, for example).

ion/tokenizer.go
```diff
@@ -1379,3 +1383,15 @@ func isNewLineChar(c int) bool {
 	return c == 0x0A || //new line
 		c == 0x0D //carriage return
 }
+
+func isClobWithInvalidChar(clob bool, c int) bool {
```
zslayton, line 1387: Throughout: I suggest using `is____` for `bool` variable names.

ion/skipper.go
```diff
@@ -804,6 +804,12 @@ func stopForCommentsHandler() (bool, error) {
 	return false, nil
 }
 
+// ensureNoCommentsHandler is a commentHandler that returns an
+// error if any comments are found, else no error is returned.
+func (t *tokenizer) ensureNoCommentsHandler() (bool, error) {
+	return false, &UnexpectedTokenError{"No Comments are allowed", t.Pos() - 1}
```
zslayton, line 810: This error message might be confusing without some additional context:

```suggestion
	return false, &UnexpectedTokenError{"Comments are not allowed within long-form strings.", t.Pos() - 1}
```

ion/integration_test.go
```diff
@@ -109,18 +109,6 @@ var malformedIonsSkipList = []string{
...
-	"clobWithNonAsciiCharacter.ion",
-	"clobWithNonAsciiCharacterMultiline.ion",
-	"clobWithValidUtf8ButNonAsciiCharacter.ion",
-	"clob_2.ion",
-	"clob_U0000003F.ion",
-	"clob_U00000080.ion",
-	"clob_U0000013F.ion",
-	"clob_u0020.ion",
-	"clob_u00FF.ion",
-	"clob_u01FF.ion",
```
zslayton, line 123: Always nice to see this many items removed from the skip list at once! 👍

ion/tokenizer.go
```diff
@@ -356,9 +356,9 @@ func (t *tokenizer) ReadValue(tok token) (string, error) {
 	case tokenSymbolOperator, tokenDot:
 		str, err = t.readOperator()
 	case tokenString:
-		str, err = t.readString()
+		str, err = t.readString(false)
```
zslayton, line 359: Open question: This PR adds an unlabeled boolean to several methods; is there a way we can make the meaning function call parameters like this more obvious?

Is it worth making `readClobString`, `readLongClobString`, etc? I suspect the answer is no, but you have more context here than I do.

ion/tokenizer.go
Armanbqt: We have already done that in an earlier [pull request](https://github.com/amzn/ion-go/pull/58/commits/460b748701e995b6754e8ad29726b98a375644c8); it is [```isProhibitedControlChar()```](https://github.com/amzn/ion-go/blob/master/ion/tokenizer.go#L1361) function.

Both of these checks are done on both [```readString()```](https://github.com/amzn/ion-go/pull/130/files#diff-99145fa564e42a4a98e734a9e12a7c55R586) and [```readLongString()```](https://github.com/amzn/ion-go/pull/130/files#diff-99145fa564e42a4a98e734a9e12a7c55R630).

I did not combine them into one function, as my understanding was ```string``` can hold characters beyond ```\x7f``` - Unless I am wrong. I do not see any mention of 7-bit ASCII limitation for string in the doc and also looked at some of the string [examples in good directory](https://github.com/amzn/ion-tests/blob/master/iontestdata/good/strings.ion#L9) where they contain 8-bit characters.

ion/skipper.go
Armanbqt: Isn't it the case that we _can have_ comments in long-form string?
```
( '''hello '''     // Sexp with one element
  '''world!'''  )
```
And the restriction is only on clob, I assume.
How about "Comments are not allowed within text format clob." ?

ion/tokenizer.go
zslayton: > We have already done that in an earlier pull request; it is isProhibitedControlChar() function.

Thanks for pointing that out. PR reviews always involve looking at the codebase through a keyhole -- only the items in the diff are brought to my attention. 

I'd still change the name of `isClobWithInvalidChar`. It's really checking to see if the current character is ASCII or not, which is a subset of checking for clob character validity. Maybe `isClobWithNonAsciiChar`?

> I did not combine them into one function, as my understanding was string can hold characters beyond \x7f - Unless I am wrong.

No, you're right. The `string` type holds unicode code points and so can store non-ascii character data. I don't think you need to combine the two functions, I just think the `isClobWithInvalidChar` function should have a name which describes its purpose more crisply.

ion/tokenizer.go
Armanbqt: > is there a way we can make the meaning function call parameters like this more obvious?

How about:
```
const isClob = true
const isNotClob = false
```
And using these two where I need to distinguish clobs from string:   ```str, err = t.readString(isNotClob)```


> Is it worth making readClobString, readLongClobString, etc?

I think the majority of what is being done in these functions are similar between clob and string, except some small changes.

ion/tokenizer.go
Armanbqt: Will update the name to ```isClobWithNonAsciiChar```

I think we can close #130?

ion/tokenizer.go
Armanbqt: Would you rather just cutting ```is``` prefix from all the following functions:
```
isProhibitedControlChar()
isStringWhitespace() 
isNewLineChar()  
isClobWithNonAsciiChar()
```
(All these methods are added either in this PR or #58)

ion/skipper.go
fernomac: (Extreme nitpick: per https://github.com/golang/go/wiki/CodeReviewComments#error-strings error strings in go should not start with a capitalized word or end in punctuation.)

ion/tokenizer.go
```diff
@@ -583,7 +583,7 @@ func (t *tokenizer) readString() (string, error) {
 			return "", err
 		}
 		// -1 denotes EOF, and new lines are not allowed in short string
-		if c == -1 || c == '\n' || isProhibitedControlChar(c) {
+		if c == -1 || c == '\n' || isProhibitedControlChar(c) || isClobWithInvalidChar(clob, c) {
```
fernomac, line 586: `... || (clob && !isASCII(c))` would read more naturally to me here 🤷.

ion/tokenizer.go
zslayton: > I think we can close #130?

#130 is this PR. Did you mean to post a different number?

Armanbqt: I meant this issue: https://github.com/amzn/ion-docs/issues/130

ion/skipper.go
zslayton: > Isn't it the case that we _can have_ comments in long-form string?
> 
> ```
> ( '''hello '''     // Sexp with one element
>   '''world!'''  )
> ```
> 
> And the restriction is only on clob, I assume.
> How about "Comments are not allowed within text format clob." ?

Yes, that's correct. Since the binary format doesn't have comments, I think you can simplify it further to `comments are not allowed within a clob`.

ion/tokenizer.go
Armanbqt: Oh, sorry. I just noticed the issue is opened in ion-doc repo

zslayton: Oh! It's the same issue number in a different repo. What a fun coincidence.

> Oh, sorry. I just noticed the issue is opened in ion-doc repo

Yeah, the `ion-docs` issue is reorganize that section of the spec so the "non-displayable" restriction on clob ASCII characters is more easily discoverable.

ion/tokenizer.go
zslayton: I'm suggesting changing the boolean variable `clob` to `isClob` for readability. Removing the `is` from those functions seems to be going in the opposite direction?

ion/tokenizer.go
zslayton: > `... || (clob && !isASCII(c))` would read more naturally to me here 🤷.

This would work for me too.

Armanbqt: Will change.

ion/tokenizer.go
Armanbqt: I thought you were talking about keeping ```is___``` only for variable names and not for function names.

I will updated the ```clob``` variable to ```isClob``` - I was following the existing ```clob``` variable.

ion/tokenizer.go
zslayton: > ```go
> const isClob = true
> const isNotClob = false
> ```

This is a good approach. My only quibble with it is that `isClob` and `isNotClob` sound like open-ended boolean values rather than constants with well-known values. For example, if I read this:

```go
str, err = t.readString(isClob)
```

it seems like I'll need to track down the definition of `isClob` to see whether it's `true` or `false`. How about:

```go
const clobText = true;
const nonClobText = false;
```

The downside is you have booleans that don't "look like" booleans (they don't start with `is`), but the upside is that the function calls end up being more self-documenting:

```go
str, err = t.readString(clobText);
// or
str, err = t.readString(nonClobText);
```

Thoughts?

Armanbqt: Yeah, I think ```clobText``` and ```nonClobText``` are better options, specially given we will name our input variable ```isClob```.

Will update the PR.

Issue #48, opened 2020-05-29
Using 'string' only to refer to text throughout the APIs


Armanbqt:
Golang allows you to use the string data type to mean "arbitrary bytes", but we should not use it that way.
We should make sure we're only using string to refer to text throughout the APIs.
[Mentioned here](https://github.com/amzn/ion-go/pull/33#discussion_r432534080)

Armanbqt:
I went through the code and didn't see any other instance of this case.

@zslayton, please let me know if you want to double check this or If I can close it.

PR #139, opened 2020-08-13
Handle clobs with escape sequence made up of 7-bit ASCII characters

### LINKED ISSUE #134: Finish up clob support
Implement failing test cases for clobs

Armanbqt:
Resolves #134 

Text format Clob is limited to 7-bit ascii characters. To make clob values transcoding between binary and text format, we have to escape non 7-bit ascii characters in text format.

Some more explanation in [this comment](https://github.com/amzn/ion-tests/pull/27#issuecomment-299272704) on ion-test repository.

### DIFF
--- a/ion/tokenizer.go
+++ b/ion/tokenizer.go
@@ -616,7 +616,7 @@ func (t *tokenizer) readString(isClob bool) (string, error) {
 			if err != nil {
 				return "", err
 			}
-			ret.WriteRune(r)
+			writeCharToStringBuilder(r, &ret, isClob)
 
 		default:
 			ret.WriteByte(byte(c))
@@ -671,7 +671,7 @@ func (t *tokenizer) readLongString(isClob bool) (string, error) {
 			if err != nil {
 				return "", err
 			}
-			ret.WriteRune(r)
+			writeCharToStringBuilder(r, &ret, isClob)
 
 		default:
 			ret.WriteByte(byte(c))
@@ -1401,4 +1401,12 @@ func getLongStringCommentsHandler(t *tokenizer, isClob bool) commentHandler {
 		return t.ensureNoCommentsHandler
 	}
 	return t.skipCommentsHandler
+}
+
+func writeCharToStringBuilder(c rune, sb *strings.Builder, isClob bool) {
+	if isClob {
+		sb.WriteByte(byte(c))
+	} else {
+		sb.WriteRune(c)
+	}
 }
--- a/ion/tokenizer.go
+++ b/ion/tokenizer.go
@@ -1403,10 +1403,11 @@ func getLongStringCommentsHandler(t *tokenizer, isClob bool) commentHandler {
 	return t.skipCommentsHandler
 }
 
-// If isClob is false, we write the UTF-8 encoding of rune, into the string builder.
-// But when isClob is true, we have a text clob. In this case, we write the bytes of
-// current rune. Clob cannot hold non 7-bit ASCII characters and to avoid truncating
-// a rune beyond 127, we store the bytes in the string builder.
+// If `isClob` is false, we write the UTF-8 encoding of `c` into the string builder. When
+// `writeCharToStringBuilder` returns, `sb` will contain a valid string.
+// If `isClob` is true, `c` is a single byte of an unknown encoding. We will write that byte to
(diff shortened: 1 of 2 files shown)


ion/tokenizer.go
```diff
@@ -1402,3 +1402,11 @@ func getLongStringCommentsHandler(t *tokenizer, isClob bool) commentHandler {
 	}
 	return t.skipCommentsHandler
 }
+
+func writeCharToStringBuilder(c rune, sb *strings.Builder, isClob bool) {
+	if isClob {
+		sb.WriteByte(byte(c))
```
fernomac, line 1408: Ahhhhh. This threw me for quite a loop at first, but I think it's correct and I think it's ultimately my fault that it's confusing. :) Explaining what I came up with to check if it matches your understanding:

When `isClob` is true the "string" we're building here is _actually_ the raw unescaped bytes of the clob, which are not necessarily a valid UTF-8 string. (The caller guarantees that `c <= 0xFF` if isClob is true, so the cast to `byte` won't truncate it, but it might be >= 0x80 which WriteRune would UTF8-icize.) We're storing those bytes in a `strings.Builder` to reuse the logic for reading and unescaping _actual_ Ion text strings, but they're not real strings and `textReader.onLob` just grabs the underlying bytes. It'd be more clear to have a separate codepath for clobs that used a `bytes.Buffer` instead of a `strings.Builder` but that would duplicate a lot of code. (Or that's my after-the-fact justification for having done it this way originally, which might be wrong? How gross would it be to refactor?)

If we do keep things this way, it might be worth noting some of this in a comment here. :)

Armanbqt: > The caller guarantees that c <= 0xFF if isClob is true, so the cast to byte won't truncate it, but it might be >= 0x80 which WriteRune would UTF8-icize

Yes, that sums up why I did this change.

> It'd be more clear to have a separate codepath for clobs that used a bytes.Buffer instead of a strings.Builder but that would duplicate a lot of code. 

It was brought up by @zslayton too in [a previous pull request](https://github.com/amzn/ion-go/pull/130#discussion_r469420291) around clobs, to which I argued code redundancy as well - of course it was before this new situation.

Thoughts @zslayton?

zslayton: The amount of code we need to duplicate does feel pretty substantial; I'd be ok with some explanatory comments. I'll leave it to @Armanbqt's discretion.

ion/tokenizer.go
```diff
@@ -1403,6 +1403,10 @@ func getLongStringCommentsHandler(t *tokenizer, isClob bool) commentHandler {
 	return t.skipCommentsHandler
 }
 
+// If isClob is false, we write the UTF-8 encoding of rune, into the string builder.
+// But when isClob is true, we have a text clob. In this case, we write the bytes of
+// current rune. Clob cannot hold non 7-bit ASCII characters and to avoid truncating
+// a rune beyond 127, we store the bytes in the string builder.
```
zslayton, line 1409: Making a note for future cleanup: I think that this code path is an instance of issue #48. We're using a string builder as a byte buffer; this works but is pretty confusing for readers. The end result isn't required to be a valid string.

> In this case, we write the bytes of current rune.

This is a bit muddy to me; when `isClob` is `true`, the current rune, `c`, is guaranteed to be a single byte of an unknown encoding.

> Clob cannot hold non 7-bit ASCII characters and to avoid truncating a rune beyond 127, we store the bytes in the string builder.

The text clob we're reading from can only hold 7-bit ASCII characters, but at this point in the code that's no longer relevant. We've already read the next byte (escaped or otherwise) from that clob; someone reading this function's documentation will be concerned with the contents of `c`, not its original format.

```go
// If `isClob` is false, we write the UTF-8 encoding of `c` into the string builder. When 
// `writeCharToStringBuilder` returns, `sb` will contain a valid string.
// If `isClob` is true, `c` is a single byte of an unknown encoding. We will write that byte to 
// the string builder as-is. When `writeCharToStringBuilder` returns, the contents of `sb` 
// must be treated as a byte array of unknown encoding.
```

Now that I've spent some more time with this, I do think we should refactor it. The cognitive burden is pretty high. It doesn't need to block this merge, though. I'll open a ticket to track it.

PR #146, opened 2020-08-20
Fix failing Annotations tests

justing-bq:


ion/binaryreader.go
```diff
@@ -443,9 +443,10 @@ func (r *binaryReader) readAnnotations() error {
 		return err
 	}
 
-	as := make([]string, len(ids))
+	as := make([]SymbolToken, len(ids))
 	for i, id := range ids {
-		as[i] = r.resolve(id)
+		name := r.resolve(id)
+		as[i] = SymbolToken{Text: &name, LocalSID: SymbolIDUnknown}
```
fernomac, line 449: `LocalSID: id`?

ion/binarywriter.go
```diff
@@ -492,17 +492,19 @@ func (w *binaryWriter) beginValue(api string) error {
...
+		var ids []uint64
 		idlen := uint64(0)
 
-		for i, a := range as {
-			id, err := w.resolve(api, a)
-			if err != nil {
-				return err
-			}
+		for _, a := range as {
+			if a.Text != nil {
```
fernomac, line 499: Is silently dropping annotations with `Text: nil` the correct behavior here? Feels like we should either be writing out a.LocalSID or failing?

ion/textwriter.go
```diff
@@ -394,11 +394,13 @@ func (w *textWriter) writeAnnotations() error {
 	w.annotations = nil
 
 	for _, a := range as {
-		if err := writeSymbol(a, w.out); err != nil {
-			return err
-		}
-		if err := writeRawString("::", w.out); err != nil {
-			return err
+		if a.Text != nil {
```
fernomac, line 397: Similarly: should we either write out `${LocalSID}` or kick back an error instead of silently skipping?

ion/binaryreader.go
justing-bq: Addressed.

ion/binarywriter.go
justing-bq: Addressed.

ion/textwriter.go
justing-bq: Addressed.

ion/binarywriter.go
```diff
@@ -495,10 +495,16 @@ func (w *binaryWriter) beginValue(api string) error {
...
-			id, err := w.resolve(api, a)
-			if err != nil {
-				return err
+			if a.Text != nil {
+				id, err = w.resolve(api, *a.Text)
+				if err != nil {
+					return err
+				}
+			} else {
+				id = 0
```
zslayton, line 507: Is this intended to support pass-through for Ion data containing unresolveable symbols?

I believe this also allows a user to write:
```go
writer.Annotation(nil)
```
and produce a `$0` in the output. That should be an error instead.

Down the road, when we offer a `AnnotationSymbolToken(SymbolToken)` method alongside the `Annotation(string)` method, we would want this to use the `LocalSID` field's value if it was set (i.e. `!= -1`) instead of using `0`.

ion/cmp_test.go
```diff
@@ -51,8 +51,28 @@ func (thisTimestamp ionTimestamp) eq(other ionEqual) bool {
 	return false
 }
 
-func cmpAnnotations(thisAnnotations, otherAnnotations []string) bool {
-	return reflect.DeepEqual(thisAnnotations, otherAnnotations)
+func cmpAnnotations(thisItem, otherItem ionItem) bool {
+	// Annotation sets are considered equal if the first annotation for each is $ion_symbol_table.
+	// eg. $ion_symbol_table::foo and $ion_symbol_table::bar::baz are considered equal.
```
zslayton, line 56: Why? Just for the purposes of this test suite, or is this from the Ion spec?

ion/cmp_test.go
```diff
@@ -51,8 +51,28 @@ func (thisTimestamp ionTimestamp) eq(other ionEqual) bool {
...
+
+	// If the first annotation is $ion_symbol_table, then we want the annotation comparison logic to also apply to
+	// all the inner structs within the ion item.
+	if !thisItem.insideLocalSymbolTable &&
+		len(thisItem.annotations) > 0 && thisItem.annotations[0] == "$ion_symbol_table" {
+		thisItem.setInsideLocalSymbolTable(true)
```
zslayton, line 62: It's surprising to me that a method with the signature `cmp(this, other) bool` has stateful side effects.

ion/reader.go
```diff
@@ -227,7 +227,21 @@ func (r *reader) FieldName() *string {
...
+	for _, an := range r.annotations {
+		if an.Text != nil {
+			systemSymbolName := getSystemSymbolMapping(V1SystemSymbolTable, *an.Text)
+			if systemSymbolName != "" {
+				annotations = append(annotations, systemSymbolName)
+			} else {
+				annotations = append(annotations, *an.Text)
+			}
+		} else {
+			annotations = append(annotations, "")
```
zslayton, line 240: This code block uses the empty string as a sentinel value in several places, which worries me. Do we need to use `* string` instead of `string` to communicate the possibility of a value's absence throughout?

ion/symboltoken.go
```diff
@@ -97,14 +100,38 @@ func (st *SymbolToken) Equal(o *SymbolToken) bool {
...
+	// then we want to use the corresponding system symbol name.
+	if len(symbolName) > 1 && symbolName[0] == '$' {
+		if id, err := strconv.Atoi(symbolName[1:]); err == nil {
+			if systemSymbolName, ok := symbolTable.FindByID(uint64(id)); ok {
+				return systemSymbolName
+			}
+		}
+	}
+
+	return ""
```
zslayton, line 114: While it is not the case today, it is possible for a system symbol table to contain the empty string.

ion/symboltoken.go
```diff
@@ -97,14 +100,38 @@ func (st *SymbolToken) Equal(o *SymbolToken) bool {
...
 // NewSymbolToken will check and return a symbol token if it exists in a symbol table,
 // otherwise return a new symbol token.
 func NewSymbolToken(symbolTable SymbolTable, text string) SymbolToken {
+	if symbolTable == nil {
+		symbolTable = V1SystemSymbolTable
+	}
+
+	systemSymbolName := getSystemSymbolMapping(symbolTable, text)
+	if systemSymbolName != "" {
+		return SymbolToken{Text: &systemSymbolName, LocalSID: SymbolIDUnknown}
```
zslayton, line 126: If the system symbol table contains the text in question, why is the `LocalSID` unknown?

ion/textwriter.go
```diff
@@ -393,8 +393,15 @@ func (w *textWriter) writeAnnotations() error {
 	as := w.annotations
 	w.annotations = nil
 
+	var text string
 	for _, a := range as {
-		if err := writeSymbol(a, w.out); err != nil {
+		if a.Text != nil {
+			text = *a.Text
+		} else {
+			text = "$0"
```
zslayton, line 401: Similar thoughts here about quietly writing out a `$0`.

ion/cmp_test.go
justing-bq: The comments in `ion-tests/iontestdata/good/equivs/localSymbolTableWithAnnotations.ion` and `ion-tests/iontestdata/good/non-equivs/localSymbolTableWithAnnotations.ion` indicated this.

ion/binaryreader.go
```diff
@@ -443,9 +443,10 @@ func (r *binaryReader) readAnnotations() error {
 		return err
 	}
 
-	as := make([]string, len(ids))
+	as := make([]SymbolToken, len(ids))
```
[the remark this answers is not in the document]
justing-bq: Addressed.

ion/cmp_test.go
justing-bq: Reworked this to perform the side effects in readCurrentValue() instead.

ion/symboltoken.go
justing-bq: Addressed.

ion/symboltoken.go
justing-bq: Addressed.

ion/reader.go
justing-bq: I wondered the same thing but I thought we didn't want to change the reader.Annotations API?

ion/binarywriter.go
justing-bq: id is now set to the `LocalSID` if it's not -1 (and returns an error is it is -1).

ion/textwriter.go
justing-bq: text is now set to `$id` unless id is -1 (then it returns an error).

ion/cmp_test.go
zslayton: These comments are [paraphrased from the spec](http://amzn.github.io/ion-docs/docs/symbols.html#annotating-local-symbol-tables); they clarify how the presence of extra annotations does or does not affect symbol table processing. I do not believe that they impact how annotation set equivalence should be evaluated.

-----

The [comment in `equivs/localSymbolTableWithAnnotations.ion`](https://github.com/amzn/ion-tests/blob/master/iontestdata/good/equivs/localSymbolTableWithAnnotations.ion#L1-L3) says: 

> // Local symbol table structs may have additional annotations as long as the $ion_symbol_table annotation comes first.
> // Additionally, values within the local symbols struct may be annotated.

The [comment in `non-equivs/localSymbolTableWithAnnotations.ion`](https://github.com/amzn/ion-tests/blob/034503caa56683f086edb1cc05a1b6207c1b18af/iontestdata/good/non-equivs/localSymbolTableWithAnnotations.ion#L1-L2) says:

> // If local symbol table structs have additional annotations, $ion_symbol_table must be the first. 
> // Otherwise, the struct is treated as a user value.

For an example, consider this struct. A reader would know to process it as a symbol table because it is annotated with `$ion_symbol_table`.
```js
$ion_symbol_table::{
    symbols: ["foo", "bar", "baz"]
}
```
You can add other annotations to that struct. If `$ion_symbol_table` is the first annotation, the struct will still be processed as a symbol table. Otherwise, `$ion_symbol_table` just becomes another annotation on a non-symbol-table user-level struct.
```js
// This is still a symbol table even though it has the 'quux' annotation
$ion_symbol_table::quux::{
    symbols: ["foo", "bar", "baz"]
}
// This is not a symbol table.
quux::$ion_symbol_table::{
    symbols: ["foo", "bar", "baz"]
}
```

-----

All of that said, I don't believe this statement from the PR is true:

> `$ion_symbol_table::foo` and `$ion_symbol_table::bar::baz` are considered equal

ion/cmp_test.go
zslayton: 👍

ion/reader.go
justing-bq: Opened https://github.com/amzn/ion-go/issues/151.

therapon: Have we considered the option of an empty array instead of a 1-element array with the element being the empty string?

ion/cmp_test.go
justing-bq: In order to pass the equivalence tests, we need the following to be considered equal.

```js
$ion_symbol_table::{
    symbols:["foo", "bar"]
}
```
```js
$ion_symbol_table::annotated::{
    symbols:["foo", "bar"]
}
```
```js
$ion_symbol_table::annotated::{
    symbols:abc::["foo", def::"bar"]
}
```

So `$ion_symbol_table` and `$ion_symbol_table::annotated` need to pass the annotation comparison even though they are internally represented as `[]string{"$ion_symbol_table"}` and `[]string{"$ion_symbol_table", "annotated"}` respectively.
Likewise `symbols:["foo", "bar"]` and `symbols:abc::["foo", def::"bar"]` need to pass the comparison because they are inside a `$ion_symbol_table` annotated struct.

My solution is to bypass the full DeepEqual comparison between annotation sets whenever we're dealing with this `$ion_symbol_table` special case.

Do you have an issue with my solution or rather with my comments explaining the special case?

ion/reader.go
zslayton: > Have we considered the option of an empty array instead of a 1-element array with the element being the empty string?

How would you distinguish between an annotation we couldn't resolve and the absence of an annotation?

ion/cmp_test.go
zslayton: > In order to pass the equivalence tests, we need the following to be considered equal.
> 
> ```
> $ion_symbol_table::{
>     symbols:["foo", "bar"]
> }
> 
> $ion_symbol_table::annotated::{
>     symbols:["foo", "bar"]
> }
> 
> $ion_symbol_table::annotated::{
>     symbols:abc::["foo", def::"bar"]
> }
> ```

This isn't quite right. Those system-level values aren't being directly tested for equivalence at all; we're actually comparing [the user-level symbol values](https://github.com/amzn/ion-tests/blob/034503caa56683f086edb1cc05a1b6207c1b18af/iontestdata/good/equivs/localSymbolTableWithAnnotations.ion#L11) that you've omitted.

As long as the `$10` and `$11` in each of those embedded streams resolve to the same symbols, we don't care what the serialized symbol tables looked like; we don't need to compare them to each other at all.

Comparing serialized symbol table tables in and of itself feels a bit dangerous. How would this code handle LST appends, where there are [multiple serialized `$ion_symbol_table` structs](https://github.com/amzn/ion-tests/blob/034503caa56683f086edb1cc05a1b6207c1b18af/iontestdata/good/equivs/localSymbolTableAppend.ion#L18-L39) that collectively result in an equivalent symbol table?

justing-bq: Turns out that https://github.com/amzn/ion-go/pull/149 actually resolves all of this rendering my solution unnecessary. I removed all my logic for this problem from this PR and told Byron to remove `localSymbolTableWithAnnotations.ion` from the skip lists in his PR since his PR fixes this scenario.

zslayton:
Could you merge + resolve conflicts? Then I'll do a final review.

ion/symboltoken.go
```diff
@@ -89,22 +92,75 @@ func (st *SymbolToken) Equal(o *SymbolToken) bool {
 	return false
 }
 
+// Parses text of the form '$n' for some integer n.
```
zslayton, line 95: I think I remember seeing similar code in #149, which has been merged. We should consolidate these methods later if they're fulfilling the same purpose.

justing-bq: That similar code has already been consolidated in this PR.

zslayton (approved):

PR #147, opened 2020-08-22
empty annotation

Armanbqt:
Annotation wrappers with no annotation subfields should be rejected.
```emptyAnnotatedInt.10n```: \xe0\x01\x00\xea\xe3\x80\x21\x01

### DIFF
--- a/ion/bitstream.go
+++ b/ion/bitstream.go
@@ -406,6 +406,12 @@ func (b *bitstream) ReadAnnotationIDs() ([]uint64, error) {
 		// annotation container.
 		return nil, &SyntaxError{"malformed annotation", b.pos - lengthOfLength}
 	}
+	if lengthValue == 0 {
+		// Annotation wrapper with an annot_length subfield value of zero is illegal.
+		// At least one annotation must exist.
+		return nil, &SyntaxError{"malformed annotation - at least one annotation must exist",
+			b.pos - lengthOfLength}
+	}
 
 	var as []uint64
 	for lengthValue > 0 {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: ion/bitstream.go [hunk 1]
```diff
@@ -406,6 +406,12 @@ func (b *bitstream) ReadAnnotationIDs() ([]uint64, error) {
 		// annotation container.
 		return nil, &SyntaxError{"malformed annotation", b.pos - lengthOfLength}
 	}
+	if lengthValue == 0 {
+		// Annotation wrapper with an annot_length subfield value of zero is illegal.
+		// At least one annotation must exist.
+		return nil, &SyntaxError{"malformed annotation - at least one annotation must exist",
```
zslayton, line 412: Nit: when I read "this annotation doesn't exist," it makes me think I've used a symbol ID (an annotation) that doesn't map to any text (doesn't exist). Dunno if others would read that the same way, but I suggest using "must be present" or "must be specified" to avoid that error message ambiguity.

```suggestion
		// At least one annotation must be present.
		return nil, &SyntaxError{"malformed annotation: at least one annotation must be specified",
```

justing-bq: Since Arman will be gone for a while I included this PR (with your suggested change) in https://github.com/amzn/ion-go/pull/150.

zslayton: Thanks. I'll close this PR out in favor of that one.

zslayton:
Closing in favor of #150, which includes this change.

PR #149, opened 2020-08-24
Incorporate symboltoken to readers WIP

byronlin-bq:
Currently there are some integration tests failing, I am investigating

Issue #, if available:

Description of changes:


ion/binaryreader.go
```diff
@@ -212,7 +213,12 @@ func (r *binaryReader) next() (bool, error) {
 
 		// If it's a local symbol table, install it and keep going.
 		if r.ctx.peek() == ctxAtTopLevel && isIonSymbolTable(r.annotations) {
-			err := r.readLocalSymbolTable()
+
+			st, err := readLocalSymbolTable(r, r.cat)
+			if err == nil {
```
matthewh-BQ, line 218: I might be misunderstanding something. This conditional is `true` when there is no error present? We'd then set `r.lst = st` and `return `false, nil`.

Should this conditional follow the same error handling pattern as above? `err != null`, then this conditional is `true` if an error is detected and then `return false, err`.? Otherwise, `r.lst = st` and `return true, nil`?

byronlin-bq: > Otherwise, r.lst = st and return true, nil?

When we import a local symbol table, we will continue consuming the next data in the stream incase there are more local symbol tables to import, therefore we return `false,nil`

ion/readerlocalsymboltable_test.go
```diff
@@ -0,0 +1,189 @@
+package ion
+
+import (
+	"testing"
+)
+
+func TestLocalSymbolTableAppend(t *testing.T) {
+	text := "$ion_symbol_table::" +
```
fernomac, line 8: Pro go tip: backtick-delimited ("raw") string literals are great for this kind of thing. See [here](https://github.com/amzn/ion-go/blob/master/ion/textwriter_test.go#L403) for an example.

ion/binaryreader_test.go
```diff
@@ -16,13 +16,171 @@
...
 	"math"
 	"math/big"
 	"testing"
 	"time"
 )
 
+const binaryFilePath = "testdata/binary/"
+
+func TestReadBinarySingleSymbol(t *testing.T) {
+	fileBytes := loadFile(t, binaryFilePath+"single_symbol.bindat")
```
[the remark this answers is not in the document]
fernomac: `gofmt` has some [Opinions(tm)](https://github.com/golang/go/issues/12720) about this that have never made much sense to me, but are very solidly established.

(Any particular reason for loading this test data from a separate file instead of inlining it like the other tests here?)

ion/binaryreader_test.go
```diff
@@ -16,13 +16,171 @@
...
+	assertSingleSymbolTest(t, r)
+}
+
+func TestReadBinaryFieldNameSymbolAndSymbolValues(t *testing.T) {
+	table1 := NewSharedSymbolTable("table1", 1, []string{"s1", "s2"})
+	table2 := NewSharedSymbolTable("table2", 1, []string{"s3", "s4"})
+	catalog := NewCatalog(table1, table2)
+
+	buf := bytes.Buffer{}
+	bw := NewBinaryWriter(&buf, table1, table2)
```
fernomac, line 41: I generally tried to stay away from using the writers to test the readers and vice versa, lest there be a bug in both that round-trips data correctly but doesn't serialize it correctly per the spec. Could we generate this bit of test data once, make sure it looks correct, and then inline it?

ion/binaryreader_test.go
```diff
@@ -16,13 +16,171 @@
...
+	fns, err := r.FieldNameSymbol()
+	if err != nil {
+		t.Fatal("unable to get field name symbol")
+	}
+	if fns.Text != nil {
+		t.Fatal("field name symbol text is incorrect")
+	}
+	if fns.LocalSID != 10 {
+		t.Fatal("field name symbol sid is incorrect")
+	}
```
fernomac, line 92: Creating versions of `_symbolAF` and friends that deal in SymbolTokens might make the intent of these sections more clear.

ion/textreader.go
```diff
@@ -285,6 +294,16 @@ func (t *textReader) nextBeforeTypeAnnotations() (bool, error) {
 		t.state = trsBeforeContainer
 		t.valueType = StructType
 		t.value = StructType
+
+		if isIonSymbolTable(t.Annotations()) {
```
fernomac, line 298: Should we also check if we're at the top level?

ion/textreader.go
```diff
@@ -389,6 +408,23 @@ func (t *textReader) verifyUnquotedSymbol(val string, ctx string) error {
 	return nil
 }
 
+// IsSid checks if the text is a symbol identifier.
+func isSid(text string) bool {
```
fernomac, line 412: Can you use `isSymbolRef` here?

ion/reader.go
```diff
@@ -163,6 +163,17 @@ type Reader interface {
...
+	// FieldNameSymbol returns the field name associated with the current value as a SymbolToken.
+	// It returns an error if the current value has no field name.
+	FieldNameSymbol() (SymbolToken, error)
+
+	// SymbolValue returns the SymbolToken associated with the current value. It returns an
+	// error if the current value is not an Ion symbol.
+	SymbolValue() (SymbolToken, error)
+
+	// Clear clears the current value from the reader.
+	clear()
```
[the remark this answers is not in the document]
byronlin-bq: I tried removing clear() and [TestReadNullLST](https://github.com/amzn/ion-go/blob/master/ion/binaryreader_test.go#L60-L69) test failed. This is because it is checking for [empty annotations.](https://github.com/amzn/ion-go/blob/master/ion/textreader_test.go#L820-L822)

I updated clear to be exportable

ion/binaryreader.go
```diff
@@ -153,7 +154,7 @@ func (r *binaryReader) next() (bool, error) {
 			if err != nil {
 				return false, err
 			}
-			r.value = r.resolve(id)
+			r.value = id
```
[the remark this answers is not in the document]
byronlin-bq: Updated

ion/textreader_test.go
```diff
@@ -844,3 +866,37 @@ func _eof(t *testing.T, r Reader) {
 		t.Fatal(r.Err())
 	}
 }
+
+func assertZeroSymbolTest(t *testing.T, r Reader) {
```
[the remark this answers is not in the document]
byronlin-bq: Removed for now

ion/binaryreader_test.go
```diff
@@ -16,13 +16,171 @@
...
+	r.Next()
+	err = r.StepIn()
+	if err != nil {
+		t.Fatal("unable to step in")
+	}
+
+	r.Next()
+	currentType := r.Type()
+	if currentType != SymbolType {
+		t.Fatal("current type is not symbol.")
```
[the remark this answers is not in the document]
byronlin-bq: Removed

ion/binaryreader_test.go
```diff
@@ -16,13 +16,171 @@
...
+		t.Fatal("symbol value is incorrect")
+	}
+	if st.LocalSID != int64(11) {
+		t.Fatal("symbol value sid is incorrect")
+	}
+
+	r.Next()
+	currentType = r.Type()
+	if currentType != SymbolType {
+		t.Fatal("current type is not symbol.")
```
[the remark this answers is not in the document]
byronlin-bq: Removed

ion/binaryreader_test.go
```diff
@@ -16,13 +16,171 @@
...
+
+	st, err = r.SymbolValue()
+	if err != nil {
+		t.Fatal("symbol token is incorrect")
+	}
+	if st.LocalSID != int64(13) {
+		t.Fatal("symbol token sid is incorrect")
+	}
+
+	//make sure that a reader with the correct imports can read it
```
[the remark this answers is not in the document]
byronlin-bq: Updated

ion/readerlocalsymboltable_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+	r := NewReaderString(text)
+	r.Next()
+	st := r.SymbolTable()
+
+	imports := st.Imports()
+	systemTable := imports[0]
+	systemMaxID := systemTable.MaxID()
+
+	checkSymbol(t, "s1", systemMaxID+1, st)
```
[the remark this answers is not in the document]
byronlin-bq: gofmt removes the spacing between appending statements

ion/textreader.go
```diff
@@ -389,6 +408,23 @@ func (t *textReader) verifyUnquotedSymbol(val string, ctx string) error {
...
+	if string(text[0]) == "$" {
+		if len(text) > 1 {
+			for i := 1; i < len(text); i++ {
+				// check if value is a digit.
```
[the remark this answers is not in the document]
byronlin-bq: Updated

ion/binaryreader.go
```diff
@@ -501,3 +336,53 @@ func (r *binaryReader) StepOut() error {
...
+// StringValue returns the current value as a string.
+func (r *binaryReader) StringValue() (string, error) {
+	if r.valueType != StringType && r.valueType != SymbolType {
+		return "", &UsageError{"Reader.StringValue", "value is not a string"}
+	}
+	if r.value == nil {
+		return "", nil
+	}
+
+	// check if value is symbol(in the form of SID) or string.
```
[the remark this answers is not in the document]
byronlin-bq: Updated

ion/reader.go
```diff
@@ -199,10 +210,11 @@ type reader struct {
 	eof bool
 	err error
 
-	fieldName   *string
-	annotations []string
-	valueType   Type
-	value       interface{}
+	fieldName    *string
+	fieldnameSID int64
```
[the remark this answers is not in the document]
byronlin-bq: Updated in this [commit](https://github.com/amzn/ion-go/pull/149/commits/9a87cc8c89c1a04102c2d577c06c32e5c191ae8a)

ion/textreader.go
byronlin-bq: Updated

ion/textreader.go
byronlin-bq: Added top level check

ion/reader.go
```diff
@@ -372,17 +384,6 @@ func (r *reader) TimestampValue() (Timestamp, error) {
 	return r.value.(Timestamp), nil
 }
 
-// StringValue returns the current value as a string.
-func (r *reader) StringValue() (string, error) {
```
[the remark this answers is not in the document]
byronlin-bq: Updated

byronlin-bq:
> gofmt has some Opinions(tm) about this that have never made much sense to me, but are very solidly established.

(Any particular reason for loading this test data from a separate file instead of inlining it like the other tests here?)

Removed and added test that has inline ion data

ion/writer.go
```diff
@@ -164,6 +168,24 @@ func (w *writer) FieldName(val string) error {
 	return nil
 }
 
+// FieldNameSymbol sets the field name symbol for the next value written.
+// It may only be called while writing a struct.
+func (w *writer) FieldNameSymbol(val SymbolToken) error {
+	if !w.IsInStruct() {
+		w.err = errors.New("ion: Writer.FieldNameSymbol called when not writing a struct")
```
justing-bq, line 175: errors.New? 
Is there a suitable error type in err.go we can use instead? If not can we create a new error type in err.go?

ion/readerlocalsymboltable_test.go
byronlin-bq: Thanks! , updated

ion/binaryreader_test.go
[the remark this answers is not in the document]
byronlin-bq: Removed test because it is covered by [TestBinaryReadSymbols](https://github.com/amzn/ion-go/blob/35232d755b0e08ebc28448694beec4801889c483/ion/binaryreader_test.go#L421)

ion/writer.go
byronlin-bq: Updated to UsageError

ion/binaryreader_test.go
byronlin-bq: Modified _symbolAF to support SymbolToken validation. Also added _nextF to support FieldNameSymbol validation

ion/binaryreader_test.go
byronlin-bq: Removed this test and instead created `TestReadBinaryFieldNames` and `TestReadBinarySymbols` tests that takes in inline data

ion/reader.go
```diff
@@ -163,6 +163,17 @@ type Reader interface {
...
 	// IsInStruct indicates if the reader is currently positioned in a struct.
 	IsInStruct() bool
+
+	// FieldNameSymbol returns the field name associated with the current value as a SymbolToken.
+	// It returns an error if the current value has no field name.
+	FieldNameSymbol() (SymbolToken, error)
+
+	// SymbolValue returns the SymbolToken associated with the current value. It returns an
+	// error if the current value is not an Ion symbol.
+	SymbolValue() (SymbolToken, error)
```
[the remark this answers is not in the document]
byronlin-bq: Would like to get opinions from @therapon, @zslayton

ion/textreader.go
```diff
@@ -183,6 +187,16 @@ func (t *textReader) nextBeforeFieldName() (bool, error) {
 			}
 		}
 
+		if tok == tokenSymbolQuoted {
+			t.fieldNameSymbol = SymbolToken{Text: &val, LocalSID: SymbolIDUnknown}
```
zslayton, line 191: It's hard to tell at a glance without an IDE: is the value of `SymbolIDUnknown` zero?

ion/textreader.go
```diff
@@ -425,6 +459,31 @@ func (t *textReader) onSymbol(val string, tok token, ws bool) error {
 	return nil
 }
 
+// readSymbol reads a text and returns a symbol token.
```
zslayton, line 462: Could you flesh out this comment a bit?

ion/textutils.go
```diff
@@ -319,3 +319,7 @@ func parseInt(str string, radix int) (interface{}, error) {
 func parseTimestamp(val string) (Timestamp, error) {
 	return ParseTimestampFromStr(val)
 }
+
+func isStringValueNil(errMsg string) bool {
+	return errMsg == "ion: usage error in Reader.StringValue: string value is unknown"
```
zslayton, line 324: I think this is a strong sign that we have an API that should be returning `*string` instead of `string`. Requiring users to do a string comparison against an error message that could change without notice is very brittle.

ion/binaryreader.go
```diff
@@ -248,192 +259,34 @@ func (r *binaryReader) readBVM() error {
...
-func (r *binaryReader) readImport() (SharedSymbolTable, error) {
-	if r.Type() != StructType || r.IsNull() {
-		return nil, nil
-	}
-	if err := r.StepIn(); err != nil {
-		return nil, err
+// ReadSymbol reads an ID and returns a symbol token.
+func (r *binaryReader) readSymbol(id uint64) (SymbolToken, error) {
+	if id > r.SymbolTable().MaxID() {
+		return symbolTokenUndefined, &UsageError{"Reader.Next", "sid is out of range "}
```
therapon, line 281: I would add the value for `id` and `MaxID()` in the error message

ion/binaryreader_test.go
```diff
@@ -401,6 +381,36 @@ func TestReadBinaryFloats(t *testing.T) {
...
+		0x86, 0x71, 0x03, // imports: $ion_symbol_table
+		0x87, 0xB4, // symbols:[
+		0x83, 'b', 'a', 'z', // "baz" ]}
+		0x71, 0x0B, // bar
+		0x71, 0x0C, // baz
+		0x71, 0x0C, // $12
+		0x71, 0x6F, // $111
+	})
+
+	_symbolAF(t, r, nil, nil, "", SymbolToken{Text: nil, LocalSID: 11}, true, false, false)
```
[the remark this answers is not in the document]
fernomac: The AF suffix means this function accepts the expected [a]nnotations and [f]ield name as parameters, as opposed to _symbol which asserts that the value has neither. The slang implication was a happy coincidence. :) It should probably be changed or at least documented.

ion/reader.go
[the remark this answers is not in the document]
therapon: We made a similar decision before. I vote for moving these to `*SymbolToken`

ion/textreader.go
therapon: From the code this looks like a function that can take a SID as a Go `string` and use it as a SID, i.e., `readSymbol("$42")`. If that is indeed the intent, maybe call this `readAsSID`?

ion/binaryreader_test.go
```diff
@@ -301,25 +268,38 @@ func TestReadBinaryStrings(t *testing.T) {
...
+		0x81, 0x21, 0x01, // $ion: 1
+		0xEE, 0x21, 0x01, // foo: 1
+		0xEF, 0x21, 0x01, // bar: 1
+		0xF1, 0x21, 0x01, // $113: 1
+		// }
+	})
+	r.Next()
+	r.StepIn()
+	_nextF(t, r, nil, SymbolToken{Text: nil, LocalSID: 0}, false, false)
+	_nextF(t, r, newString("$ion"), SymbolToken{Text: newString("$ion"), LocalSID: 1}, false, false)
```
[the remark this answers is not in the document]
byronlin-bq: It came from this [issue](https://github.com/amzn/ion-go/issues/11) and [PR](https://github.com/amzn/ion-go/pull/14)

ion/textreader.go
byronlin-bq: `SymbolIDUnknown` has value `-1`

ion/reader.go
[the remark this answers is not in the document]
byronlin-bq: Updated to return `(*SymbolToken, error)`

ion/binaryreader.go
```diff
@@ -248,192 +259,34 @@ func (r *binaryReader) readBVM() error {
...
-		case "symbols":
-			syms, err = r.readSymbols()
-		}
-		if err != nil {
-			return err
-		}
-	}
-
-	if err := r.StepOut(); err != nil {
+	st, err := r.readSymbol(id)
```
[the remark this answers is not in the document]
byronlin-bq: Updated to `resolveSymbolID `

ion/binaryreader_test.go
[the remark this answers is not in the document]
byronlin-bq: Added docstring to `_symbolAF` function

ion/reader.go
```diff
@@ -406,3 +406,51 @@ func (r *reader) clear() {
...
 }
+
+// StringValue returns the current value as a string.
+func (r *reader) StringValue() (string, error) {
+	if r.err != nil {
+		return "", r.err
+	}
+
+	if r.valueType != StringType && r.valueType != SymbolType {
+		return "", &UsageError{"Reader.StringValue", "value is not a string"}
```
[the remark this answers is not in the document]
byronlin-bq: Updated

ion/reader.go
```diff
@@ -406,3 +406,51 @@ func (r *reader) clear() {
...
+	}
+
+	if r.value == nil {
+		return "", &UsageError{"Reader.StringValue", "string value is unknown"}
```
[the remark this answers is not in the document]
byronlin-bq: Updated stringValue() to return `(*string, error)`

ion/reader.go
```diff
@@ -406,3 +406,51 @@ func (r *reader) clear() {
...
+	}
+
+	// Check if value is symbol or string.
+	st, ok := r.value.(SymbolToken)
+	if !ok {
+		return r.value.(string), nil
+	}
+
+	if st.Text == nil {
+		return "", &UsageError{"Reader.StringValue", "string value is unknown"}
```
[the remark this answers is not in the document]
byronlin-bq: No-op because I updated stringValue() to return `(*string, error)`

ion/textreader.go
byronlin-bq: > From the code this looks like a function that can take a SID as a Go string and use it as a SID, i.e., readSymbol("$42"). If that is indeed the intent, maybe call this readAsSID?

The function reads SID text as well as symbol text e.g., `foo` and return a symboltoken.

Updated comment to:
```
// readSymbol reads a text as either a symbol identifier e.g., "$42" or a symbol text e.g., "foo" and resolves to a
// symbol token.
```

ion/textutils.go
byronlin-bq: Updated stringvalue() to return `(*string, error)`

ion/binaryreader.go
byronlin-bq: Updated

cmd/ion-go/eventwriter.go
```diff
@@ -41,6 +41,11 @@ func NewEventWriter(out io.Writer) ion.Writer {
 	return &eventwriter{enc: ion.NewEncoder(w)}
 }
 
+func (e *eventwriter) FieldNameSymbol(val ion.SymbolToken) error {
+	e.fieldname = val.Text
```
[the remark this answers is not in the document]
byronlin-bq: I added `FieldNameSymbol()` to the writer interface so I can test the integration tests.

Updated to:
```func (e *eventwriter) FieldNameSymbol(val ion.SymbolToken) error {
	if val.Text == nil {
		sid := fmt.Sprintf("$%v", val.LocalSID)
		e.fieldname = &sid
		return nil
	}

	e.fieldname = val.Text
	return nil
}```

File: ion/reader.go (hunk 8 above)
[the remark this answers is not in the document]
Comment (zslayton) [reply ion/reader.go]: I don't think we want `Clear()` to be part of the public API that users can implement. It feels like an implementation detail to me; no user would ever call `Clear()`.

In OO languages, you might have a `Writer` interface and an `AbstractWriter` base class to hold common implementation details like `Clear()`. I'm unsure what the best way to model that in Go would be.

File: ion/unmarshal_test.go [hunk 32]
```diff
@@ -340,11 +340,11 @@ func TestUnmarshalBinary(t *testing.T) {
 
 	var symbolVal string
 	symbolBytes := prefixIVM([]byte{0x71, 0x09}) // $9
-	test(symbolBytes, symbolVal, "$ion_shared_symbol_table")
+	test(symbolBytes, symbolVal, newString("$ion_shared_symbol_table"))
 
 	var stringVal string
 	stringBytes := prefixIVM([]byte{0x83, 'a', 'b', 'c'}) // "abc"
-	test(stringBytes, stringVal, "abc")
+	test(stringBytes, stringVal, newString("abc"))
```
zslayton, line 347: I'm not sure I understand the value of [the `newString` function](https://github.com/amzn/ion-go/blob/58d1539a051c268383698a3566e3ef176d994c1e/ion/symboltoken_test.go#L25-L27):

```go
func newString(value string) *string {
	return &value
}
```

It's certainly correct, but why is `newString("hello")` preferable to `&"hello"`?

ion/integration_test.go
```diff
@@ -689,8 +680,9 @@ func readCurrentValue(t *testing.T, reader Reader) ionItem {
 	}
 
 	currentType := reader.Type()
-	if reader.IsNull() {
-		ionItem.value = append(ionItem.value, textNulls[currentType])
+	// Excluding SymbolType from setting value to 'null.symbol' because readers can read "'null.symbol'" as text: 'null.symbol'.
```
zslayton, line 683: > because readers can read "'null.symbol'" as text: 'null.symbol'

I don't think I understand. If a text reader encounters a `null.symbol`, it has an ion type of `symbol`, `IsNull()` is `true`, and `StringValue()` would return `nil`. 

How would a reader read it as text?

ion/reader.go
```diff
@@ -155,14 +155,25 @@ type Reader interface {
 
 	// StringValue returns the current value as a string (if that makes sense). It returns
 	// an error if the current value is not an Ion symbol or an Ion string.
-	StringValue() (string, error)
+	StringValue() (*string, error)
```
zslayton, line 158: This really simplifies [the case from my earlier comment](https://github.com/amzn/ion-go/pull/149#discussion_r481272560); if the string or symbol was `null`, this function will return a `nil` that can be passed around. Great.

However, we now have a larger consistency question for the `Reader` interface. Should all values return `*type` so they return a `nil` if the Ion value is `null`? At the moment most of them return an `err` with the expectation that you will have handled `null` separately through a combination of `IsNull()` and `Type()`. (This isn't documented yet, by the way. See [`IntValue()` for an example](https://github.com/amzn/ion-go/blob/1f421dfa4f274ec841989150ea325434170aac0f/ion/reader.go#L125-L128) -- the doc comment doesn't explain what to expect if the value is `null`.)

Having two layers of checking to do after calling each `*Value()` function (Is it an `err`? Is it `nil`?) feels pretty heavy to me, which is why I think we were generating a `UsageError` for `null` values originally.

This is a larger discussion, but it's something I think we need to get right before 1.0. I'm ok with opening an issue to track this so it doesn't block the PR.

ion/unmarshal_test.go
fernomac: Because only one of them successfully compiles? :) It’s a bit of a chicken-and-egg: it’s not very “go-ey” to use pointers to basic types in APIs, so there isn’t a lot of built-in syntax sugar to make working with them easy (in this case: you can’t directly take the address of a literal, you need to assign it to a variable and then take the address of the variable), which reinforces the convention against using them in APIs.

ion/integration_test.go
byronlin-bq: This stems from this [test](https://github.com/amzn/ion-tests/blob/master/iontestdata/good/non-equivs/symbols.ion#L25-L26).

According to the symbol specs, if a text reader encounters a single quoted text, it has an ion type of symbol, IsNull() is false, and StringValue() would return the text.

In this case, StringValue() will return `null.symbol` which is the same value appended by the test when text reader encounters unquoted text `null.symbol`

ion/readlocalsymboltable.go
```diff
@@ -0,0 +1,206 @@
...
+
+	var imps []SharedSymbolTable
+	var syms []string
+
+	foundImport := false
+	foundLocals := false
+
+	for r.Next() {
+		var err error
+		switch *r.FieldName() {
```
justing-bq, line 24: What if r.FieldName() is nil? Let's add an error case for that.

ion/reader.go
byronlin-bq: Created ticket: https://github.com/amzn/ion-go/issues/152

ion/integration_test.go
zslayton: > According to the symbol specs, if a text reader encounters a single quoted text, it has an ion type of symbol, IsNull() is false, and StringValue() would return the text.

Agreed. I think my confusion stems from the comment's phrasing:

> because readers **can read** "'null.symbol'" as text: 'null.symbol'.

If a `null.symbol` is wrapped in single quotes (`'null.symbol'`), it *must* be read as text, and never as a null.

> In this case, StringValue() will return "null.symbol" which is the same value appended by the test when text reader encounters unquoted text null.symbol

I would expect `reader.StringValue()` to return:
* `"null.symbol"` when the Ion value is `'null.symbol'` (a symbol)
* An `err` (or `nil`, depending on #152) when the Ion value is `null.symbol` (a null)

Is that what you're describing, or am I missing something?

byronlin-bq: > Is that what you're describing, or am I missing something?

That is correct

ion/unmarshal_test.go
zslayton: > Because only one of them successfully compiles? :)

That's a good reason. :)

> you can’t directly take the address of a literal, you need to assign it to a variable and then take the address of the variable

Huh.

justing-bq: This isn't a choice between `newString("hello")` and `&"hello"`. 
Rather it's a choice between `newString("hello")` and 
```go
str := "hello"
&str
```

ion/reader.go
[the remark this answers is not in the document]
byronlin-bq: Removed `Clear()` from `Reader` interface. `Clear()` was being used in the beginning of [readLocalSymbolTable()](https://github.com/amzn/ion-go/pull/149/commits/851ab03c6a785fb196e6722d01ed0be28ed4ae9f#diff-c7a62f6b5becd9a9a65eb57ed6c40180L7-L10) and I move that logic to to text/binary readers just before they call `readLocalSymbolTable`

ion/readlocalsymboltable.go
byronlin-bq: Added a check for nil fieldname

ion/unmarshal_test.go
zslayton: I understood what @fernomac was describing, I was just surprised to learn that that was a restriction in Go.

zslayton (approved):

therapon (approved):

PR #150, opened 2020-08-26
Fix malformed Ion annotations tests

justing-bq:


ion/bitstream.go
```diff
@@ -396,26 +396,45 @@ func (b *bitstream) ReadAnnotationIDs() ([]uint64, error) {
...
 			return nil, err
 		}
 
 		as = append(as, id)
-		lengthValue -= idlen
+		annotFieldLength -= idlen
+	}
+
+	// Get the raw data for the value enclosed within the annotation without advancing the reader.
+	annotatedData, err := b.in.Peek(int(remainingAnnotationLength))
```
fernomac, line 430: This could plausibly be quite large. Is there any way to do the needed checks without reading the whole annotated value into memory at once?

justing-bq: Addressed. Now we're just peeking at the next byte one at a time.

PR #153, opened 2020-09-03
Implement AnnotationsAsSymbols

justing-bq:

Dependent on https://github.com/amzn/ion-go/pull/146

### DIFF
--- a/ion/reader.go
+++ b/ion/reader.go
@@ -103,6 +103,9 @@ type Reader interface {
 	// It returns nil if there is no current value or the current value has no annotations.
 	Annotations() []string
 
+	// AnnotationsAsSymbols returns the set of annotations associated with the current value as SymbolTokens.
+	AnnotationsAsSymbols() []SymbolToken
+
 	// StepIn steps in to the current value if it is a container. It returns an error if there
 	// is no current value or if the value is not a container. On success, the Reader is
 	// positioned before the first value in the container.
@@ -238,6 +241,11 @@ func (r *reader) FieldName() *string {
 
 // Annotations returns the current value's annotations.
 func (r *reader) Annotations() []string {
+	return r.annotations
+}
+
+// AnnotationsAsSymbols returns the current value's annotations as SymbolTokens.
+func (r *reader) AnnotationsAsSymbols() []SymbolToken {
 	return r.annotations
 }
 
--- a/ion/symboltoken.go
+++ b/ion/symboltoken.go
@@ -92,19 +92,6 @@ func (st *SymbolToken) Equal(o *SymbolToken) bool {
 	return false
 }
 
-// ResolveToString returns a SymbolToken's text value or "$<sid>"
-func (st *SymbolToken) ResolveToString() string {
-	if st.Text != nil {
-		return *st.Text
-	}
-
-	if st.LocalSID != SymbolIDUnknown {
-		return fmt.Sprintf("$%v", st.LocalSID)
-	}
-
-	return ""
-}
-
 // Parses text of the form '$n' for some integer n.
 func symbolIdentifier(symbolText string) (int64, bool) {
 	if len(symbolText) > 1 && symbolText[0] == '$' {
--- a/ion/symboltoken.go
+++ b/ion/symboltoken.go
@@ -89,6 +89,19 @@ func (st *SymbolToken) Equal(o *SymbolToken) bool {
 	return false
 }
 
+// ResolveToString returns a SymbolToken's text value or "$<sid>"
+func (st *SymbolToken) ResolveToString() string {
+	if st.Text != nil {
+		return *st.Text
+	}
+
+	if st.LocalSID != SymbolIDUnknown {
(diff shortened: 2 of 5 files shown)


ion/reader.go
```diff
@@ -98,6 +98,9 @@ type Reader interface {
 	// It returns nil if there is no current value or the current value has no annotations.
 	Annotations() []string
```
zslayton, line 99: I was under the impression that we'd be replacing this method (`Annotations() []string`) instead of adding a new one alongside it since `[]string` makes error reporting challenging.

ion/reader.go
```diff
@@ -98,6 +98,9 @@ type Reader interface {
...
 
+	// AnnotationsAsSymbols returns the set of annotations associated with the current value as SymbolTokens.
+	AnnotationsAsSymbols() []SymbolToken
```
zslayton, line 102: Where does validation for symbol IDs being out of bounds take place?

If a `Reader` encounters this value when the symbol table's MaxId is `100`:
```js
$33::$0::$705::"foo"
```
what will be returned by a call to `AnnotationsAsSymbols`? Does the user get three `SymbolToken` instances and it's up to them to detect that the third one is illegal? I think I'd expect a `([]SymbolToken, error)` signature to handle that case, but want to make sure I understand what you had in mind.

ion/writer.go
```diff
@@ -78,6 +78,12 @@ type Writer interface {
 	// Annotations adds multiple annotations to the next value written.
 	Annotations(values ...string) error
```
zslayton, line 79: If we're leaving the `string`-centric `Annotations` method here: are `string`s that aren't in the symbol table quietly added to it before the value is written out, or will it produce an `error`? Either way, that behavior should be clarified in the documentation.

ion/writer.go
```diff
@@ -78,6 +78,12 @@ type Writer interface {
...
 
+	// Annotation adds a single annotation as a SymbolToken to the next value written.
+	AnnotationAsSymbol(val SymbolToken) error
+
+	// Annotations adds multiple annotations as SymbolTokens to the next value written.
+	AnnotationsAsSymbols(values ...SymbolToken) error
```
zslayton, line 85: The documentation should indicate whether passing a `SymbolToken` that's not already in the table is an `error` or not.

ion/reader.go
justing-bq: So `r.AnnotationsAsSymbols()` simply returns the reader's internal `r.Annotations` slice of type `[]SymbolToken`.

The symbol ID validation happens in the `SymbolToken` constructor during `r.Next()`. As a result the invalid `SymbolToken` never gets appended to the reader's `r.annotations` slice.

ion/symboltoken.go
```diff
@@ -92,6 +92,19 @@ func (st *SymbolToken) Equal(o *SymbolToken) bool {
 	return false
 }
 
+// ResolveToString returns a SymbolToken's text value or "$<sid>"
+func (st *SymbolToken) ResolveToString() string {
```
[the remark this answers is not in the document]
justing-bq: The motivation behind this function is for getting some ion-hash tests passing in https://github.com/amzn/ion-hash-go/pull/62.

```go
val, err := reader.SymbolValue()
assert.NoError(t, err, "Something went wrong when reading Symbol value")

assert.NoError(t, writer.WriteSymbol(val.ResolveToString()), "Something went wrong when writing Symbol value")
```

Note that `writer.go` currently has `WriteSymbol(val string)` not `WriteSymbol(val SymbolToken)`.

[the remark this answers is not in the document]
zslayton: > The motivation behind this function is for getting some ion-hash tests passing 

My gut reaction is that this method is not serving a common use case outside of our own libraries and so shouldn't be part of the public API we commit to in v1.0. Could/should this just be a private helper function in the `ion-hash-go` test suite?

> Note that writer.go currently has WriteSymbol(val string) not WriteSymbol(val SymbolToken).

I thought the takeaway from last week's meeting was that we'd be using `SymbolToken` in all of the `Reader`/`Writer` APIs that involve symbols so we'd have a consistent (if a bit tedious) error reporting story.

ion/reader.go
zslayton: So the error would be surfaced when the user called `r.Next()` and the `r.annotations` slice is being populated, preventing `r.AnnotationsAsSymbols()` from ever being called?

ion/symboltoken.go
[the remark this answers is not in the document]
justing-bq: I took your suggestion and moved the ResolveToString() logic to ion-hash-go. After making that change the rest of this PR is basically pointless since the annotations API is being redone by Byron anyway.

I will close this PR and let Byron's PR handle the annotations API changes.

justing-bq:
Closed PR because annotations API changes are being overhauled in https://github.com/byronlin-bq/ion-go/pull/6.

PR #154, opened 2020-09-09
[Refactor] Make a single object for all different types acceptable kind

### LINKED ISSUE #110: Improve performance on validating structs' fields type for unmarshalling values with annotation
As pointed out in [this review comment](https://github.com/amzn/ion-go/pull/99#discussion_r461255083_):

> At the moment we're allocating a new slice for [type]AcceptableKinds in every decode[Type]To method and we're doing a linear search over those kinds in this method.
I think we can make a single object to hold maps of the acceptable kinds for each type so we can allocate them once and search them quickly.

Armanbqt:
Resolves #110 

Define and allocate acceptable kinds for different Ion types once, to optimize the existing logic.

### DIFF
--- a/ion/unmarshal.go
+++ b/ion/unmarshal.go
@@ -30,6 +30,14 @@ var (
 	ErrNoInput = errors.New("ion: no input to decode")
 )
 
+var typesAcceptableKinds = map[Type][]reflect.Kind{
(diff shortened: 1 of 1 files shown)


zslayton (approved):
Nice!

PR #155, opened 2020-09-09
Handle Negative Zero and NaN Float

justing-bq:
ion-hash-go needs these changes to get float tests passing for -0e0 and NaN.


ion/binarywriter.go
```diff
@@ -191,6 +192,11 @@ func (w *binaryWriter) WriteFloat(val float64) error {
 	bits := math.Float64bits(val)
 	binary.BigEndian.PutUint64(bs[1:], bits)
 
+	// The above PutUint64() puts an unwanted 1 as the last byte for NaN.
```
Armanbqt, line 195: Can you confirm ```NaN``` is always 8 bits? If ```NaN``` can be presented in 4 bits (float32) we might want to add the same logic on line 185 too.

justing-bq: We know the last index is 8 because we have `bs := make([]byte, 9)` above.

Armanbqt: On [line 179](https://github.com/amzn/ion-go/pull/155/files#diff-2e0c44cac09fc56c07997526b83a087eR179) we check whether the value is a float32 or 64. If it is a float32 then we'll have ```make([]byte, 5)``` otherwise array byte of size 9.
My concern was if there is a case where ```NaN``` is float32

justing-bq: Here the float32 case doesn't happen for NaN because we're getting a float64 value passed in and float64 NaN does not pass the `val == float64(float32(val))` condition.

Regardless I checked anyway and discovered that the unwanted 1 issue only applies to the 64 bit case.

Armanbqt: Thanks

therapon (approved):

ion/binarywriter.go
```diff
@@ -170,26 +170,32 @@ func (w *binaryWriter) writeBigInt(val *big.Int) error {
...
-
-	bs := make([]byte, 9)
-	bs[0] = 0x48
+		bits := math.Float64bits(val)
+		binary.BigEndian.PutUint64(bs[1:], bits)
 
-	bits := math.Float64bits(val)
-	binary.BigEndian.PutUint64(bs[1:], bits)
+		// The above PutUint64() puts an unwanted 1 as the last byte for NaN.
+		if math.IsNaN(val) {
```
zslayton, line 195: If we handle `NaN` as a special case right after `-0`, we can avoid having to allocate a buffer and set/unset that byte.

```go
if val == 0 && !math.Signbit(val) {
    // Positive zero is represented as just one byte.
    return w.writeValue("Writer.WriteFloat", []byte{0x40})
} else if math.IsNaN(val) {
    return w.writeValue("Writer.WriteFloat", []byte{/* <bytes go here> */})
}
```
This also lets us represent a binary `NaN` in 4 bytes instead of 8.

justing-bq: Addressed.

zslayton (approved):

PR #156, opened 2020-09-11
Handle Negative Zero Decimal

justing-bq:

### DIFF
--- a/ion/decimal.go
+++ b/ion/decimal.go
@@ -351,15 +351,24 @@ func (d *Decimal) String() string {
 
 	case d.scale < 0:
 		// Value is a upscaled integer, nn'd'ss
+		if d.isNegZero {
+			return "-0d" + fmt.Sprintf("%d", -d.scale)
+		}
 		return d.n.String() + "d" + fmt.Sprintf("%d", -d.scale)
 
 	default:
 		// Value is a downscaled integer nn.nn('d'-ss)?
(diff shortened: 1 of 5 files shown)


ion/decimal.go
```diff
@@ -38,21 +38,23 @@ func (e *ParseError) Error() string {
 
 // Decimal is an arbitrary-precision decimal value.
 type Decimal struct {
-	n     *big.Int
-	scale int32
+	n         *big.Int
+	scale     int32
+	isNegZero bool
```
Armanbqt, line 43: Do you think we can use ```isNegZero``` in ```cmp()``` function?

justing-bq: Addressed.

ion/integration_test.go
```diff
@@ -124,11 +124,7 @@ var equivsSkipList = []string{
 	"systemSymbols.ion",
 }
 
-var nonEquivsSkipList = []string{
-	"decimals.ion",
-	"floats.ion",
-	"floatsVsDecimals.ion",
-}
+var nonEquivsSkipList = []string{}
```
Armanbqt, line 127: ❤

justing-bq: Feels good to see an empty skiplist.

therapon (approved):

zslayton (approved):

PR #157, opened 2020-09-11
Reader api change

### LINKED ISSUE #152: Inconsistency on the reader interface when returning values.
Currently on reader API, some values that return `*type` so they return a `nil` if the Ion value is `null`. e.g., [DecimalValue](https://github.com/amzn/ion-go/blob/master/ion/reader.go#L150)

However most of them return an err with the expectation that you will have handled null separately through a combination of IsNull() and Type(). (This isn't documented yet, by the way. See IntValue() for an example -- the doc comment doesn't explain what to expect if the value is null.)

This ticket is a discussion on whether we should return `*type` for all values.

Armanbqt:
Resolves #152 

As #152 explains, this pull request is to update Reader api interface.

### DIFF
--- a/ion/integration_test.go
+++ b/ion/integration_test.go
@@ -703,21 +703,21 @@ func readCurrentValue(t *testing.T, reader Reader) ionItem {
 	}
 
 	currentType := reader.Type()
-	// Excluding SymbolType from setting value to 'null.symbol' because readers can read "'null.symbol'" as text: 'null.symbol'.
-	if reader.IsNull() && reader.Type() != SymbolType {
-		ionItem.value = append(ionItem.value, nil)
-		ionItem.ionType = currentType
-
-		return ionItem
-	}
-
 	switch currentType {
+	case NullType:
+		ionItem.value = append(ionItem.value, nil)
+		ionItem.ionType = NullType
+
 	case BoolType:
 		val, err := reader.BoolValue()
 		if err != nil {
 			t.Errorf("Something went wrong when reading Boolean value. " + err.Error())
 		}
-		ionItem.value = append(ionItem.value, *val)
+		if val == nil {
+			ionItem.value = append(ionItem.value, nil)
+		} else {
+			ionItem.value = append(ionItem.value, *val)
+		}
 		ionItem.ionType = BoolType
 
 	case IntType:
@@ -725,7 +725,11 @@ func readCurrentValue(t *testing.T, reader Reader) ionItem {
 		if err != nil {
 			t.Errorf("Something went wrong when reading Int value. " + err.Error())
 		}
-		ionItem.value = append(ionItem.value, *val)
+		if val == nil {
+			ionItem.value = append(ionItem.value, nil)
+		} else {
+			ionItem.value = append(ionItem.value, *val)
+		}
 		ionItem.ionType = IntType
 
 	case FloatType:
@@ -733,7 +737,11 @@ func readCurrentValue(t *testing.T, reader Reader) ionItem {
 		if err != nil {
 			t.Errorf("Something went wrong when reading Float value. " + err.Error())
 		}
-		ionItem.value = append(ionItem.value, *val)
+		if val == nil {
+			ionItem.value = append(ionItem.value, nil)
+		} else {
+			ionItem.value = append(ionItem.value, *val)
+		}
 		ionItem.ionType = FloatType
 
 	case DecimalType:
@@ -741,7 +749,11 @@ func readCurrentValue(t *testing.T, reader Reader) ionItem {
 		if err != nil {
 			t.Errorf("Something went wrong when reading Decimal value. " + err.Error())
 		}
-		ionItem.value = append(ionItem.value, val)
+		if val == nil {
+			ionItem.value = append(ionItem.value, nil)
+		} else {
+			ionItem.value = append(ionItem.value, val)
+		}
 		ionItem.ionType = DecimalType
 
 	case TimestampType:
@@ -749,7 +761,11 @@ func readCurrentValue(t *testing.T, reader Reader) ionItem {
 		if err != nil {
 			t.Errorf("Something went wrong when reading Timestamp value. " + err.Error())
 		}
-		ionItem.value = append(ionItem.value, *val)
+		if val == nil {
+			ionItem.value = append(ionItem.value, nil)
+		} else {
+			ionItem.value = append(ionItem.value, *val)
+		}
 		ionItem.ionType = TimestampType
 
 	case SymbolType:
@@ -765,7 +781,11 @@ func readCurrentValue(t *testing.T, reader Reader) ionItem {
 		if err != nil {
 			t.Errorf("Something went wrong when reading String value. " + err.Error())
 		}
-		ionItem.value = append(ionItem.value, *val)
+		if val == nil {
+			ionItem.value = append(ionItem.value, nil)
+		} else {
+			ionItem.value = append(ionItem.value, *val)
+		}
 		ionItem.ionType = StringType
 
 	case ClobType:
@@ -785,6 +805,11 @@ func readCurrentValue(t *testing.T, reader Reader) ionItem {
 		ionItem.ionType = BlobType
 
 	case SexpType:
+		if reader.IsNull() {
+			ionItem.value = append(ionItem.value, nil)
+			ionItem.ionType = SexpType
+			break
+		}
 		err := reader.StepIn()
 		if err != nil {
 			t.Fatal(err)
@@ -799,6 +824,11 @@ func readCurrentValue(t *testing.T, reader Reader) ionItem {
 		}
 
 	case ListType:
+		if reader.IsNull() {
+			ionItem.value = append(ionItem.value, nil)
+			ionItem.ionType = ListType
+			break
+		}
 		err := reader.StepIn()
 		if err != nil {
 			t.Fatal(err)
@@ -813,6 +843,11 @@ func readCurrentValue(t *testing.T, reader Reader) ionItem {
 		}
 
 	case StructType:
+		if reader.IsNull() {
+			ionItem.value = append(ionItem.value, nil)
+			ionItem.ionType = StructType
+			break
+		}
 		err := reader.StepIn()
 		if err != nil {
 			t.Fatal(err)
--- a/ion/integration_test.go
+++ b/ion/integration_test.go
@@ -463,14 +463,27 @@ func writeFromReaderToWriter(t *testing.T, reader Reader, writer Writer) {
 		}
 
 		switch currentType {
+		case NullType:
+			err := writer.WriteNullType(currentType)
+			if err != nil {
+				t.Errorf("Something went wrong while writing a Null value: " + err.Error())
+			}
+
 		case BoolType:
 			val, err := reader.BoolValue()
 			if err != nil {
 				t.Errorf("Something went wrong while reading a Boolean value: " + err.Error())
 			}
-			err = writer.WriteBool(*val)
-			if err != nil {
-				t.Errorf("Something went wrong while writing a Boolean value: " + err.Error())
+			if val == nil {
+				err := writer.WriteNullType(BoolType)
+				if err != nil {
+					t.Fatal(err)
+				}
+			} else {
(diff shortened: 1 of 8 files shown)


ion/integration_test.go
```diff
@@ -673,7 +705,7 @@ func readCurrentValue(t *testing.T, reader Reader) ionItem {
 	currentType := reader.Type()
 	// Excluding SymbolType from setting value to 'null.symbol' because readers can read "'null.symbol'" as text: 'null.symbol'.
 	if reader.IsNull() && reader.Type() != SymbolType {
```
byronlin-bq, line 707: we can remove `reader.Type() != SymbolType` and add 

```
			if val == nil {
				err := writer.WriteNullType(SymbolType)
				if err != nil {
					t.Fatal(err)
				}
```

under `case SymbolType`

byronlin-bq: Do we even need this `reader.IsNull()` conditional? if a value is nil, the value APIs will return a `nil`

Armanbqt: At this point we still have not read the value. This is to avoid going through the switch statement; as soon as we get to a null value we create corresponding ```ionItem``` for it.

Can you explain the code snippet please? Why should it be ```WriteNullType(SymbolType)```, what if we are reading ```null.timestamp```?

byronlin-bq: > At this point we still have not read the value. This is to avoid going through the switch statement; as soon as we get to a null value we create corresponding ionItem for it.

I think it would be good idea if we let it go through the switch statement, to test that the value APIs do return `nil` instead of the test setting the value to `nil`

> Can you explain the code snippet please? Why should it be WriteNullType(SymbolType), what if we are reading null.timestamp?

My bad, I thought the reason why we were excluding symbolType being set to `nil` was because the writer wasn't writing a `null` symbol correctly. Disregard the comment

Armanbqt: Done

ion/integration_test.go
```diff
@@ -671,53 +703,69 @@ func readCurrentValue(t *testing.T, reader Reader) ionItem {
...
 		val, err := reader.BoolValue()
 		if err != nil {
 			t.Errorf("Something went wrong when reading Boolean value. " + err.Error())
 		}
-		ionItem.value = append(ionItem.value, val)
+		if val == nil {
+			ionItem.value = append(ionItem.value, nil)
+		} else {
+			ionItem.value = append(ionItem.value, *val)
+		}
```
therapon, line 720: We cannot really refactor this repeating code due to `val` being a different type in each `case`-branch, right? I am checking here and fear that I am right--I wish I wasn't. :)

ion/reader.go
```diff
@@ -109,47 +109,49 @@ type Reader interface {
 	// stream.
 	StepOut() error
 
-	// BoolValue returns the current value as a boolean (if that makes sense). It returns
-	// an error if the current value is not an Ion bool.
-	BoolValue() (bool, error)
+	// BoolValue returns the current value as a boolean (if that makes sense). It returns nil
+	// if the value is null. If the current value is not an Ion bool, it returns an error.
```
therapon, line 113:
```suggestion
	// if the value is an Ion null. If the current value is not an Ion bool, it returns an error.
```

I am not sure if we need to make the distinction here since Go has `nil` and not `null`.

ion/reader.go
```diff
@@ -109,47 +109,49 @@ type Reader interface {
...
 	// IntSize returns the size of integer needed to losslessly represent the current value
 	// (if that makes sense). It returns an error if the current value is not an Ion int.
 	IntSize() (IntSize, error)
 
-	// IntValue returns the current value as a 32-bit integer (if that makes sense). It
-	// returns an error if the current value is not an Ion integer or requires more than
-	// 32 bits to represent losslessly.
-	IntValue() (int, error)
+	// IntValue returns the current value as a 32-bit integer (if that makes sense). It returns
+	// nil if the value is null. It returns an error if the current value is not an Ion integer
```
therapon, line 121:
```suggestion
	// nil if the value is an Ion null. It returns an error if the current value is not an Ion integer
```

ion/reader.go
```diff
@@ -109,47 +109,49 @@ type Reader interface {
...
+	// or requires more than 32 bits to represent losslessly.
+	IntValue() (*int, error)
 
-	// Int64Value returns the current value as a 64-bit integer (if that makes sense). It
-	// returns an error if the current value is not an Ion integer or requires more than
-	// 64 bits to represent losslessly.
-	Int64Value() (int64, error)
+	// Int64Value returns the current value as a 64-bit integer (if that makes sense). It returns
+	// nil if the value is null. It returns an error if the current value is not an Ion integer
```
therapon, line 126:
```suggestion
	// nil if the value is an Ion null. It returns an error if the current value is not an Ion integer
```

ion/reader.go
```diff
@@ -109,47 +109,49 @@ type Reader interface {
...
+	// or requires more than 64 bits to represent losslessly.
+	Int64Value() (*int64, error)
 
 	// Uint64Value returns the current value as an unsigned 64-bit integer (if that makes
-	// sense). It returns an error if the current value is not an Ion integer, is negative,
-	// or requires more than 64 bits to represent losslessly.
-	Uint64Value() (uint64, error)
+	// sense). It returns nil if the value is null. It returns an error if the current value
```
therapon, line 131:
```suggestion
	// sense). It returns nil if the value is an Ion null. It returns an error if the current value
```

ion/reader.go
```diff
@@ -109,47 +109,49 @@ type Reader interface {
...
+	// is not an Ion integer, is negative, or requires more than 64 bits to represent losslessly.
+	Uint64Value() (*uint64, error)
 
-	// BigIntValue returns the current value as a big.Integer (if that makes sense). It
-	// returns an error if the current value is not an Ion integer.
+	// BigIntValue returns the current value as a big.Integer (if that makes sense). It returns
+	// nil if the value is null. It returns an error if the current value is not an Ion integer.
```
therapon, line 136:
```suggestion
	// nil if the value is an Ion null. It returns an error if the current value is not an Ion integer.
```

ion/reader.go
```diff
@@ -109,47 +109,49 @@ type Reader interface {
...
+	// FloatValue returns the current value as a 64-bit floating point number (if that makes
+	// sense). It returns nil if the value is null. It returns an error if the current value
+	// is not an Ion float.
+	FloatValue() (*float64, error)
 
-	// DecimalValue returns the current value as an arbitrary-precision Decimal (if that
-	// makes sense). It returns an error if the current value is not an Ion decimal.
+	// DecimalValue returns the current value as an arbitrary-precision Decimal (if that makes
+	// sense). It returns nil if the value is null. It returns an error if the current value is
+	// not an Ion decimal.
```
therapon, line 146: Do we have a bound on the Decimal's size that we (our code) error out on? If the error in the case of something too large is from the language that's ok. I am asking if we perform any checks on top of that and generate an error ourselves.

ion/reader.go
```diff
@@ -257,14 +259,15 @@ func (r *reader) Annotations() []string {
 }
 
 // BoolValue returns the current value as a bool.
-func (r *reader) BoolValue() (bool, error) {
+func (r *reader) BoolValue() (*bool, error) {
 	if r.valueType != BoolType {
-		return false, &UsageError{"Reader.BoolValue", "value is not a bool"}
+		return nil, &UsageError{"Reader.BoolValue", "value is not a bool"}
```
therapon, line 264:
```suggestion
		return nil, &UsageError{"Reader.BoolValue", "value is not an Ion bool"}
```

ion/reader.go
```diff
@@ -292,62 +295,66 @@ func (r *reader) IntSize() (IntSize, error) {
 }
 
 // IntValue returns the current value as an int.
-func (r *reader) IntValue() (int, error) {
+func (r *reader) IntValue() (*int, error) {
 	i, err := r.Int64Value()
```
therapon, line 299: Should we have an if-statement here to check the type and throw an appropriate error if the function is called on an incorrect Ion Type?

ion/reader.go
```diff
@@ -292,62 +295,66 @@ func (r *reader) IntSize() (IntSize, error) {
...
+	val := int(*i)
+	return &val, nil
 }
 
 // Int64Value returns the current value as an int64.
-func (r *reader) Int64Value() (int64, error) {
+func (r *reader) Int64Value() (*int64, error) {
 	if r.valueType != IntType {
-		return 0, &UsageError{"Reader.Int64Value", "value is not an int"}
+		return nil, &UsageError{"Reader.Int64Value", "value is not an int"}
```
therapon, line 313:
```suggestion
		return nil, &UsageError{"Reader.Int64Value", "value is not an Ion int"}
```

ion/reader.go
```diff
@@ -292,62 +295,66 @@ func (r *reader) IntSize() (IntSize, error) {
...
-	return 0, &UsageError{"Reader.Int64Value", "value too large for an int64"}
+	return nil, &UsageError{"Reader.Int64Value", "value too large for an int64"}
 }
 
 // Uint64Value returns the current value as a uint64.
-func (r *reader) Uint64Value() (uint64, error) {
+func (r *reader) Uint64Value() (*uint64, error) {
 	if r.valueType != IntType {
-		return 0, &UsageError{"Reader.Uint64Value", "value is not an int"}
+		return nil, &UsageError{"Reader.Uint64Value", "value is not an int"}
```
therapon, line 335:
```suggestion
		return nil, &UsageError{"Reader.Uint64Value", "value is not an Ion int"}
```

ion/reader.go
```diff
@@ -366,14 +373,15 @@ func (r *reader) BigIntValue() (*big.Int, error) {
 }
 
 // FloatValue returns the current value as a float.
-func (r *reader) FloatValue() (float64, error) {
+func (r *reader) FloatValue() (*float64, error) {
 	if r.valueType != FloatType {
-		return 0, &UsageError{"Reader.FloatValue", "value is not a float"}
+		return nil, &UsageError{"Reader.FloatValue", "value is not a float"}
```
therapon, line 378:
```suggestion
		return nil, &UsageError{"Reader.FloatValue", "value is not an Ion float"}
```

ion/reader.go
```diff
@@ -388,14 +396,15 @@ func (r *reader) DecimalValue() (*Decimal, error) {
 }
 
 // TimestampValue returns the current value as a Timestamp.
-func (r *reader) TimestampValue() (Timestamp, error) {
+func (r *reader) TimestampValue() (*Timestamp, error) {
 	if r.valueType != TimestampType {
-		return Timestamp{}, &UsageError{"Reader.TimestampValue", "value is not a timestamp"}
+		return nil, &UsageError{"Reader.TimestampValue", "value is not a timestamp"}
```
therapon, line 401:
```suggestion
		return nil, &UsageError{"Reader.TimestampValue", "value is not an Ion timestamp"}
```

ion/integration_test.go
Armanbqt: I'm afraid you are right. I unsuccessfully spent quite some time to find a more elegant way for this.

ion/reader.go
Armanbqt: ```r.Int64Value()``` here calls another function in this file (right below) [```Int64Value```](https://github.com/amzn/ion-go/pull/157/files#diff-0a0ff18f6967f2ebfe8b753560b3c836R311) where the check on type is done.

ion/reader.go
Armanbqt: Short answer is no.
bigint does not have an explicit limit.

> There is no explicit limit. The limit will be your system resources (memory) allow.

[here](https://ispycode.com/Blog/golang/2016-10/What-is-the-max-value-of-bigint) and [here](https://stackoverflow.com/questions/17564335/golang-math-big-what-is-the-max-value-of-big-int#:~:text=There%20is%20no%20explicit%20limit,where%2010%5E100000%20is%20benchmarked.) for reference.

ion/reader.go
therapon: You are right. I missed that, thanks.

therapon (approved):

zslayton (approved):

PR #165, opened 2020-10-02
Symbol Token Write Support

justing-bq:

### DIFF
--- a/cmd/ion-go/main.go
+++ b/cmd/ion-go/main.go
@@ -78,14 +78,14 @@ func printVersion() error {
 		return err
 	}
 	{
-		if err := w.FieldName(ion.NewSimpleSymbolToken("version")); err != nil {
+		if err := w.FieldName(ion.NewSymbolTokenFromString("version")); err != nil {
 			return err
 		}
 		if err := w.WriteString(internal.GitCommit); err != nil {
 			return err
 		}
 
-		if err := w.FieldName(ion.NewSimpleSymbolToken("build_time")); err != nil {
+		if err := w.FieldName(ion.NewSymbolTokenFromString("build_time")); err != nil {
 			return err
 		}
 
--- a/cmd/ion-go/main.go
+++ b/cmd/ion-go/main.go
@@ -78,28 +78,24 @@ func printVersion() error {
 		return err
 	}
 	{
-		if err := w.FieldName("version"); err != nil {
+		if err := w.FieldName(ion.NewSimpleSymbolToken("version")); err != nil {
 			return err
 		}
 		if err := w.WriteString(internal.GitCommit); err != nil {
 			return err
 		}
 
+		if err := w.FieldName(ion.NewSimpleSymbolToken("build_time")); err != nil {
+			return err
+		}
+
 		buildtime, err := ion.NewTimestampFromStr(internal.BuildTime, ion.TimestampPrecisionSecond, ion.TimezoneUTC)
 		if err == nil {
-			if err := w.FieldName("build_time"); err != nil {
-				return err
-			}
 			if err := w.WriteTimestamp(buildtime); err != nil {
 				return err
 			}
-		} else {
-			if err := w.FieldName("build_time"); err != nil {
-				return err
-			}
-			if err := w.WriteString("unknown-buildtime"); err != nil {
-				return err
-			}
+		} else if err := w.WriteString("unknown-buildtime"); err != nil {
+			return err
 		}
 	}
 	if err := w.EndStruct(); err != nil {
--- a/ion/binarywriter.go
+++ b/ion/binarywriter.go
@@ -21,8 +21,6 @@ import (
 	"io"
 	"math"
 	"math/big"
-	"strconv"
-	"strings"
 	"time"
 )
 
@@ -251,35 +249,43 @@ func (w *binaryWriter) WriteTimestamp(val Timestamp) error {
 	return w.writeValue("Writer.WriteTimestamp", buf)
 }
 
+// WriteSymbol writes a symbol value given a SymbolToken.
 func (w *binaryWriter) WriteSymbol(val SymbolToken) error {
 	var id uint64
-	if val.LocalSID == SymbolIDUnknown {
-		id, w.err = w.resolve1("Writer.WriteSymbol", *val.Text)
+	if val.LocalSID != SymbolIDUnknown {
+		id = uint64(val.LocalSID)
+	} else if val.Text != nil {
+		text := *val.Text
+		if _, ok := symbolIdentifier(text); ok {
+			// Wrap text value in single quotes if the symbol's text is a symbol identifier
+			// (ie. of form $n for some integer n)
+			// This is done to distinguish from actual symbol table mappings.
+			text = fmt.Sprintf("'%v'", text)
+		}
+
+		id, w.err = w.resolveFromSymbolTable("Writer.WriteSymbol", text)
 		if w.err != nil {
 			return w.err
 		}
 	} else {
-		id = uint64(val.LocalSID)
+		return &UsageError{"Writer.WriteSymbol", "invalid symbol token"}
 	}
 
-	vlength := uintLen(id)
-	bufLength := vlength + tagLen(vlength)
-	buf := make([]byte, 0, bufLength)
-
-	buf = appendTag(buf, 0x70, vlength)
-	buf = appendUint(buf, id)
-
-	return w.writeValue("Writer.WriteSymbol", buf)
+	return w.writeSymbolFromID("Writer.WriteSymbol", id)
 }
 
-// WriteSymbolFromString writes a symbol value.
+// WriteSymbolFromString writes a symbol value given a string.
 func (w *binaryWriter) WriteSymbolFromString(val string) error {
-	id, err := w.resolve("Writer.WriteSymbolFromString", val)
-	if err != nil {
-		w.err = err
-		return err
+	var id uint64
+	id, w.err = w.resolve("Writer.WriteSymbolFromString", val)
+	if w.err != nil {
+		return w.err
 	}
 
+	return w.writeSymbolFromID("Writer.WriteSymbolFromString", id)
+}
+
+func (w *binaryWriter) writeSymbolFromID(api string, id uint64) error {
 	vlength := uintLen(id)
 	bufLength := vlength + tagLen(vlength)
 	buf := make([]byte, 0, bufLength)
@@ -287,7 +293,7 @@ func (w *binaryWriter) WriteSymbolFromString(val string) error {
 	buf = appendTag(buf, 0x70, vlength)
 	buf = appendUint(buf, id)
 
-	return w.writeValue("Writer.WriteSymbolFromString", buf)
+	return w.writeValue(api, buf)
 }
 
 // WriteString writes a string.
@@ -618,27 +624,20 @@ func (w *binaryWriter) end(api string, t ctx) error {
 
 // Resolve resolves a symbol to its ID.
 func (w *binaryWriter) resolve(api, sym string) (uint64, error) {
-	if strings.HasPrefix(sym, "$") {
-		id, err := strconv.ParseUint(sym[1:], 10, 64)
-		if err == nil {
-			return id, nil
-		}
+	if id, ok := symbolIdentifier(sym); ok {
+		return uint64(id), nil
 	}
 
-	return w.resolve1(api, sym)
+	return w.resolveFromSymbolTable(api, sym)
 }
 
-func (w *binaryWriter) resolve1(api, sym string) (uint64, error) {
+func (w *binaryWriter) resolveFromSymbolTable(api, sym string) (uint64, error) {
 	if w.lst != nil {
 		id, ok := w.lst.FindByName(sym)
 		if !ok {
 			return 0, &UsageError{api, fmt.Sprintf("symbol '%v' not defined", sym)}
 		}
 		return id, nil
-	}
-
-	if _, ok := symbolIdentifier(sym); ok {
-		sym = fmt.Sprintf("'%v'", sym)
 	}
 
 	id, _ := w.lstb.Add(sym)
--- a/ion/binarywriter.go
+++ b/ion/binarywriter.go
@@ -255,26 +255,19 @@ func (w *binaryWriter) WriteSymbol(val SymbolToken) error {
 	if val.LocalSID != SymbolIDUnknown {
 		id = uint64(val.LocalSID)
 	} else if val.Text != nil {
-		text := *val.Text
-		if _, ok := symbolIdentifier(text); ok {
-			// Wrap text value in single quotes if the symbol's text is a symbol identifier
-			// (ie. of form $n for some integer n)
-			// This is done to distinguish from actual symbol table mappings.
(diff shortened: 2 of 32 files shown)


ion/symboltable.go
```diff
@@ -202,7 +214,7 @@ func (s *sst) String() string {
 	buf := strings.Builder{}
 
 	w := NewTextWriter(&buf)
-	s.WriteTo(w)
+	_ = s.WriteTo(w)
```
Armanbqt, line 217: I think ```WriteTo()``` returns ```error```.
Why not checking the error whether it's nil or not?

ion/symboltable.go
```diff
@@ -281,19 +293,22 @@ func (s *bogusSST) String() string {
 
 	buf := strings.Builder{}
 	w := NewTextWriter(&buf)
-	w.Annotations(SymbolToken{Text: &ionSharedSymbolTableText, LocalSID: 9}, SymbolToken{Text: &bogusText, LocalSID: SymbolIDUnknown})
```
Armanbqt, line 284: Warnings, ha!? Yikes

ion/writer.go
```diff
@@ -148,40 +146,22 @@ type writer struct {
...
-
-	w.fieldName = &val
-	return nil
-}
-
-// FieldNameSymbol sets the field name symbol for the next value written.
-// It may only be called while writing a struct.
-func (w *writer) FieldNameSymbol(val SymbolToken) error {
 	if !w.IsInStruct() {
 		w.err = &UsageError{"Writer.FieldNameSymbol", "called when not writing a struct"}
```
Armanbqt, line 160: "Writer.FieldName" instead of "Writer.FieldNameSymbol"?

ion/unmarshal.go
```diff
@@ -196,9 +199,15 @@ func (d *Decoder) decodeInt() (interface{}, error) {
...
+		}
 		return *val, err
 	case Int64:
 		val, err := d.r.Int64Value()
+		if err != nil {
+			return nil, err
+		}
 		return *val, err
 	default:
 		return d.r.BigIntValue()
```
Armanbqt, line 213: ```BigIntValue``` also returns value, error.
We can the same check as above ```cases```es for ```default``` too

ion/textwriter.go
```diff
@@ -411,21 +414,12 @@ func (w *textWriter) writeFieldName(api string) error {
...
-		if a.Text != nil {
-			text = *a.Text
-		} else if a.LocalSID != SymbolIDUnknown {
-			text = fmt.Sprintf("$%v", a.LocalSID)
-		} else {
-			return &UsageError{api, "invalid annotation symbol token"}
-		}
-
-		if err := writeSymbol(text, w.out); err != nil {
+		if err := writeSymbol(a.toText(), w.out); err != nil {
```
Armanbqt, line 422: If I read the code correctly, when we have ```a.Text``` is nil and ```a.LocalSID``` is Unknown, then ```toText()``` will return an empty string. And this empty text is sent to ```writeSymbol```.
I think if ```a.Text``` is nil and ```a.LocalSID``` is Unknown, error should be returned.

ion/symboltoken.go
```diff
@@ -103,6 +113,15 @@ func symbolIdentifier(symbolText string) (int64, bool) {
 	return SymbolIDUnknown, false
 }
 
+// NewSimpleSymbolToken returns a Symbol Token with the given text value and undefined SID and Source.
+func NewSimpleSymbolToken(text string) SymbolToken {
+	return SymbolToken{Text: &text, LocalSID: SymbolIDUnknown}
+}
+
+func newSimpleSymbolTokenPtr(text string) *SymbolToken {
```
Armanbqt, line 121: Personally, I think we should not have this. we have ```NewSimpleSymbolToken()``` and if its pointer is needed that should be sufficient.

ion/decimal.go
```diff
@@ -274,10 +272,9 @@ func (d *Decimal) trunc() (int64, error) {
 	if err != nil {
 		return 0, err
 	}
-	d = ud
```
Armanbqt, line 277: I think we need this.
```ud``` is different from the original ```d```.  I think when we change the scale, we need to keep ```d``` with its updated version in case another operation is called on ```d```

File: cmd/ion-go/main.go [hunk 8]
```diff
@@ -78,28 +78,24 @@ func printVersion() error {
...
 	{
-		if err := w.FieldName("version"); err != nil {
+		if err := w.FieldName(ion.NewSimpleSymbolToken("version")); err != nil {
 			return err
 		}
 		if err := w.WriteString(internal.GitCommit); err != nil {
 			return err
 		}
 
+		if err := w.FieldName(ion.NewSimpleSymbolToken("build_time")); err != nil {
```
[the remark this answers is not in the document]
justing-bq: Look at the old code. `w.FieldName("build_time")` is in both the nil err case and the non-nil err case.

ion/decimal.go
justing-bq: Consider `d.checkToUpscale()`. 
In the old code, it reassigns d with `d = d.upscale(0)`. However I discovered that this change of state is not reflected in the calling function `d.trunc()` which must be why `d.checkToUpscale()` returns the new decimal. (Yes I'm confused by this as well since I expect a pointer receiver to change state.)

ion/symboltable.go
justing-bq: If this were inside a function that returns an error then I would have added the error checking. But it doesn't and I'm not gonna add an error to the return type because that would block sst from satisfying the `Stringer` interface.
I added this explicit `_ =` so that Goland would no longer flag this as a warning.

ion/symboltable.go
justing-bq: Yes all of this was just to make Goland stop flagging it.

ion/symboltoken.go
justing-bq: I would love to not have this either but without it I would need two lines to get the address.

ie.
```go
st := NewSimpleSymbolToken(text)
&st
```
instead of
```go
newSimpleSymbolTokenPtr(text)
```

ion/binarywriter.go
```diff
@@ -251,22 +249,51 @@ func (w *binaryWriter) WriteTimestamp(val Timestamp) error {
...
+			return w.err
+		}
+	} else {
+		return &UsageError{"Writer.WriteSymbol", "invalid symbol token"}
+	}
+
+	return w.writeSymbolFromID("Writer.WriteSymbol", id)
+}
+
+// WriteSymbolFromString writes a symbol value given a string.
```
zslayton, line 277: This comment should be more explicit about how the provided string is going to be used; it expects the string to already be in the symbol table and will return an error if it isn't.

ion/binarywriter.go
```diff
@@ -251,22 +249,51 @@ func (w *binaryWriter) WriteTimestamp(val Timestamp) error {
...
+	var id uint64
+	if val.LocalSID != SymbolIDUnknown {
+		id = uint64(val.LocalSID)
+	} else if val.Text != nil {
+		text := *val.Text
+		if _, ok := symbolIdentifier(text); ok {
+			// Wrap text value in single quotes if the symbol's text is a symbol identifier
+			// (ie. of form $n for some integer n)
+			// This is done to distinguish from actual symbol table mappings.
+			text = fmt.Sprintf("'%v'", text)
```
zslayton, line 263: I'm not sure I understand the behavior that's intended here. Why are we wrapping `$37` in single quotes?

If I pass a `SymbolToken` with `LocalSID=SymbolIDUnknown` and `Text="$37"`, I would expect `WriteSymbol` to try to resolve the text literal `$37` (unmodified) in the symbol table. If I'd wanted to write SID 37, I would have set `LocalSID=37` in the `SymbolToken`, right?

ion/binarywriter.go
```diff
@@ -492,9 +519,17 @@ func (w *binaryWriter) beginValue(api string) error {
...
+		if name.LocalSID != SymbolIDUnknown {
+			id = uint64(name.LocalSID)
+		} else if name.Text != nil {
+			var err error
+			id, err = w.resolve(api, *name.Text)
+			if err != nil {
+				return err
+			}
+		} else {
+			return &UsageError{api, "invalid field name symbol token"}
```
zslayton, line 532: The error message should communicate why the symbol token is invalid.

ion/binarywriter.go
```diff
@@ -591,13 +626,14 @@ func (w *binaryWriter) end(api string, t ctx) error {
 
 // Resolve resolves a symbol to its ID.
 func (w *binaryWriter) resolve(api, sym string) (uint64, error) {
-	if strings.HasPrefix(sym, "$") {
-		id, err := strconv.ParseUint(sym[1:], 10, 64)
-		if err == nil {
-			return id, nil
-		}
+	if id, ok := symbolIdentifier(sym); ok {
```
zslayton, line 629: I want to understand the symbol ID interpretation behavior that we're implementing here. I'm having difficulty imagining a scenario where I'm using a binary writer and I pass it a string in the `$id` format (`$12`, `$47`, etc).

ion/textwriter.go
```diff
@@ -80,54 +80,60 @@ func (w *textWriter) WriteBool(val bool) error {
...
+		return err
 	}
-	if w.err = w.beginValue("Writer.WriteSymbol"); w.err != nil {
-		return w.err
+	if val.Text != nil {
+		// Wrap text value in single quotes if the symbol's text is a symbol identifier
+		// (ie. of form $n for some integer n)
+		// This is done to distinguish from actual symbol table mappings.
+		if _, ok := symbolIdentifier(text); ok {
+			text = fmt.Sprintf("'%v'", text)
```
zslayton, line 127: Same question here RE: quote wrapping.

ion/symboltoken.go
```diff
@@ -103,6 +112,15 @@ func symbolIdentifier(symbolText string) (int64, bool) {
 	return SymbolIDUnknown, false
 }
 
+// NewSimpleSymbolToken returns a Symbol Token with the given text value and undefined SID and Source.
+func NewSimpleSymbolToken(text string) SymbolToken {
```
zslayton, line 116: This is subjective, but to me the phrase `Simple` doesn't communicate what this function will do. How about `NewSymbolTokenFromString`?

ion/binarywriter.go
```diff
@@ -251,22 +249,51 @@ func (w *binaryWriter) WriteTimestamp(val Timestamp) error {
...
+			// This is done to distinguish from actual symbol table mappings.
+			text = fmt.Sprintf("'%v'", text)
+		}
+
+		id, w.err = w.resolveFromSymbolTable("Writer.WriteSymbol", text)
+		if w.err != nil {
+			return w.err
+		}
+	} else {
+		return &UsageError{"Writer.WriteSymbol", "invalid symbol token"}
```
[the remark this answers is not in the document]
justing-bq: Done.

ion/binarywriter.go
justing-bq: Done.

ion/binarywriter.go
justing-bq: Done.

ion/symboltoken.go
justing-bq: Done.

ion/binarywriter.go
justing-bq: We have roundtrip tests (such as `ion-tests/iontestdata/good/symbols.ion` with the test case `'$99'`) that fail without any internal logic wrapping `$99` in quotes.

When the text reader reads `'$99'`, it produces the string `"$99"` (note the lack of single quotes) and then creates a SymbolToken using that string value. The absence of single quotes creates confusion further downstream about whether we are dealing with an actual symbol table mapping vs a non-mapping SymbolToken with this text value.

I moved the single quote wrapping logic over to `textReader::nextBeforeTypeAnnotations` instead of `binaryWriter::WriteSymbol` and `textWriter::WriteSymbol`.

ion/textwriter.go
justing-bq: Addressed.

ion/writer.go
justing-bq: Addressed.

ion/textwriter.go
justing-bq: Addressed.

ion/skipper.go
```diff
@@ -546,44 +546,42 @@ func (t *tokenizer) skipLongStringHelper(handler commentHandler) error {
 // false indicating this is not the end of the long string, and true for consumed '
 // as we have read the closing ''' of the first long string.
 func (t *tokenizer) skipEndOfLongString(handler commentHandler) (bool, bool, error) {
-	isConsumed := false
```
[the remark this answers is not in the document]
justing-bq: Reverted back my change.

ion/binarywriter.go
zslayton: > such as ion-tests/iontestdata/good/symbols.ion with the test case '$99'

Here's [a link to this test case](https://github.com/amzn/ion-tests/blob/034503caa56683f086edb1cc05a1b6207c1b18af/iontestdata/good/symbols.ion#L26) for anyone who's interested.

> When the text reader reads '$99', it produces the string "$99" (note the lack of single quotes) and then creates a SymbolToken using that string value.

This sounds like the correct behavior to me.

> The absence of single quotes creates confusion further downstream about whether we are dealing with an actual symbol table mapping vs a non-mapping SymbolToken with this text value.

This is something we need to sort out; I don't think there's any writer API where the golang string `$99` should be reinterpreted as anything other than a string literal.

Can you link to the function(s) that are causing the problems? Why is the reinterpretation happening at all?

fernomac: I don't remember off the top of my head where specifically to go looking, but the root cause on this one is likely me. :)

At some point back before the introduction of `SymbolToken`, if the binary reader hit a symbol ID it couldn't resolve (or the text reader hit an unquoted ${n}), they'd return the string `${id}` instead of an error. The writers correspondingly mapped `WriteSymbol("$123")` to being a symbol with that ID, bypassing the symbol table. That came at the expense of being able to read/write an _actual_ symbol whose text is `${n}`, but that wasn't a thing I thought I would ever need to do (😬😅).

Now that everything at the reader/writer level is converted to SymbolTokens, any remnants of that questionable decision can and should be removed.

zslayton: This makes total sense. Thanks for the added context!

justing-bq: Alright I reworked this so the Symbol Token's text value is the unquoted `$99`. 
However when the text writer is writing out the symbol in textutils.go:writeSymbol(), we still need to wrap this value in single quotes.

ion/unmarshal.go
byronlin-bq, line 213: Could you update Unmarshal API to support shared symbol tables.

I was playing around marshalling/ unmarshalling roundtrip with sst and it was missing.

Update L95-97 to:
```
func Unmarshal(data []byte, v interface{}, ssts ...SharedSymbolTable ) error {
	catalog := NewCatalog(ssts...)
	return NewDecoder(NewReaderCat(bytes.NewReader(data), catalog)).DecodeTo(v)
}
```

justing-bq: Addressed.

zslayton (approved):
:shipit:

PR #167, opened 2020-10-13
implement Marshaler and Unmarshaler in ion.Decimal

whrami:
Issue #, if available: N/A

Description of changes: ion.Decimal being a complex type with private struct fields, it is unable to be marshaled to/from JSON as fields in more complex structs.

This PR implements Marshaler for the ion.Decimal receiver type, and Unmarshaler for the *ion.Decimal receiver type.

### DIFF
--- a/ion/decimal.go
+++ b/ion/decimal.go
@@ -16,6 +16,7 @@
 package ion
 
 import (
+	"encoding/json"
 	"fmt"
 	"math"
 	"math/big"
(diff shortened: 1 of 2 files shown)


ion/decimal.go
```diff
@@ -394,3 +395,26 @@ func (d *Decimal) String() string {
...
+
+// UnmarshalJSON is implementing Unmarshaler for Decimal
+func (d *Decimal) UnmarshalJSON(blob []byte) error {
+	var daj decimalAsJSON
+	if err := json.Unmarshal(blob, &daj); err != nil {
+		return fmt.Errorf("unmarshalling ion.Decimal: %v", err)
+	}
+	d.n = daj.Co
+	// CoEx() returns negated scale for exponent; re-negate it in order to restore the correct value.
+	d.scale = -daj.Ex
```
Armanbqt, line 418: Should we also check for negative zero here? ```d.isNegZero```

zslayton:
Hi, thanks for the PR!

The various Ion libraries follow [this algorithm](https://amzn.github.io/ion-docs/guides/cookbook.html#down-converting-to-json) when down-converting Ion data to JSON, which is an inherently lossy transformation. In the case of `decimal`:

> Decimals are printed as JSON number with precision preserved

A JSON `Object` is a perfectly valid representation, but that functionality would need to live in another library as it would not be consistent with our other language bindings.

PR #175, opened 2021-03-13
address a few compiler warnings

fernomac:
Description of changes:

Hey friends! I was testing out the setup of a new laptop by giving this a build and figured I'd fix a couple quick compiler warnings while I was in the neighborhood:

- Allocating an empty slice and then appending to it -> directly allocating the resulting slice
- `switch in.(type)` -> `switch in := in.(type)` to avoid a second typecheck/cast in the case statements
- casting an int to a string is discouraged because what it _actually_ does is string([]rune{rune(int)), but it looks like it's doing `strconv.Itoa(int)` (we had one of each :)).

### DIFF
--- a/ion/skipper.go
+++ b/ion/skipper.go
@@ -657,7 +657,7 @@ func (t *tokenizer) skipContainer(term int) (int, error) {
 // char.
 func (t *tokenizer) skipContainerHelper(term int) error {
 	if term != ']' && term != ')' && term != '}' {
-		panic(fmt.Sprintf("unexpected character: %v. Expected one of the closing container characters: ] } )", string(term)))
+		panic(fmt.Sprintf("unexpected character: %v. Expected one of the closing container characters: ] } )", rune(term)))
 	}
 
 	for {
--- a/ion/skipper.go
+++ b/ion/skipper.go
@@ -657,7 +657,7 @@ func (t *tokenizer) skipContainer(term int) (int, error) {
 // char.
 func (t *tokenizer) skipContainerHelper(term int) error {
 	if term != ']' && term != ')' && term != '}' {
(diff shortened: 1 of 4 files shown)


ion/skipper.go
```diff
@@ -657,7 +657,7 @@ func (t *tokenizer) skipContainer(term int) (int, error) {
 // char.
 func (t *tokenizer) skipContainerHelper(term int) error {
 	if term != ']' && term != ')' && term != '}' {
-		panic(fmt.Sprintf("unexpected character: %v. Expected one of the closing container characters: ] } )", string(term)))
+		panic(fmt.Sprintf("unexpected character: %v. Expected one of the closing container characters: ] } )", rune(term)))
```
jobarr-amzn, line 660: Is this what is wanted here? Or something more like the `%q` formatting directive with a bare `term`?

See: https://play.golang.org/p/JKRiykdCWJv

jobarr-amzn:
```plain
Using string(term):
unexpected character: A. Expected one of the closing container characters: ] } )
Using rune(term):
unexpected character: 65. Expected one of the closing container characters: ] } )
Using string(rune(term)):
unexpected character: A. Expected one of the closing container characters: ] } )
Using %c with bare term:
unexpected character: A. Expected one of the closing container characters: ] } )
Using %q with bare term:
unexpected character: 'A'. Expected one of the closing container characters: ] } )
```

fernomac: Ah, good call! :+1: for `%q`.

jobarr-amzn (commented):
Looks good to me :)

jobarr-amzn (approved):

Issue #173, opened 2020-12-17
Unexpected behavior of Timestamp fraction precision


bancek:
Amazon QLDB fails to insert a timestamp if fraction precision is `3` and milliseconds are divisible by `10`.

Example:
```go
type Entry struct {
	Time ion.Timestamp `ion:"time"`
}

millis := int64(1608218715340)
t := time.Unix(0, millis*1000000)
ts := ion.NewTimestampWithFractionalSeconds(t, ion.TimestampPrecisionNanosecond, ion.TimezoneUTC, 3)

entry := Entry{
	Time: ts,
}

_, err = driver.Execute(context.Background(), func(txn qldbdriver.Transaction) (interface{}, error) {
	return txn.Execute("INSERT INTO Test ?", entry)
})
```

The error from Amazon QLDB is
```
BadRequestException: Digests don't match
{
  RespMetadata: {
    StatusCode: 400,
    RequestID: "3GH3JFguR5O3LLcAwTQiiN"
  },
  Code_: "412",
  Message_: "Digests don't match"
}
```

Is this an expected behavior?

I've fixed this with a helper to calculate the fraction precision but I would expect the `ion-go` library to do automatically.

```go
func getFractionPrecision(t time.Time) uint8 {
	millis := (t.UnixNano() / 1000000) % 1000
	fractionPrecision := uint8(3)

	if millis == 0 {
		fractionPrecision = 0
	} else {
		for millis > 0 && (millis%10) == 0 {
			millis /= 10
			fractionPrecision--
		}
	}

	return fractionPrecision
}
```

This is also a problem when using `time.Time` in a `struct` and setting it to `time.Now().UTC()`. It fails with same error `Digests don't match`.

Full example:

```go
package main

import (
	"context"
	"time"

	"github.com/amzn/ion-go/ion"
	"github.com/aws/aws-sdk-go/aws"
	"github.com/aws/aws-sdk-go/aws/session"
	"github.com/aws/aws-sdk-go/service/qldbsession"
	"github.com/awslabs/amazon-qldb-driver-go/qldbdriver"
)

func main() {
	awsSession := session.Must(session.NewSession(aws.NewConfig().WithRegion("us-east-1")))
	qldbSession := qldbsession.New(awsSession)

	driver, err := qldbdriver.New(
		"quick-start",
		qldbSession,
		func(options *qldbdriver.DriverOptions) {
			options.LoggerVerbosity = qldbdriver.LogInfo
		})
	if err != nil {
[...]

PR #177, opened 2021-05-28
Correct the coefficient value of timestamp nanoseconds when writing to Ion binary

byronlin13:
**Issue #, if available:**
* #173 
* [error inserting ion.Timestamp values](https://github.com/awslabs/amazon-qldb-driver-go/issues/52)
* [error inserting time.Time values](https://github.com/awslabs/amazon-qldb-driver-go/issues/51)

This PR is to address digest mismatch error when passing various timestamp objects into [QLDB driver for Go](https://github.com/awslabs/amazon-qldb-driver-go).

**Timestamps tested:**
* `timestamp := time.Now()`
* `timestamp := time.Now().UTC()`
*
        millis := int64(1608218715340)
        t := time.Unix(0, millis*1000000)
        timeStamp := ion.NewTimestampWithFractionalSeconds(t, ion.TimestampPrecisionNanosecond, ion.TimezoneUTC, 3)
* `timestamp := ion.NewTimestampWithFractionalSeconds(time.Now().UTC(), ion.TimestampPrecisionNanosecond, ion.TimezoneUTC, 9)`
* `timestamp := ion.NewTimestampWithFractionalSeconds(time.Now().UTC(), ion.TimestampPrecisionSecond, ion.TimezoneUTC, 9)`
*  `timestamp, _ := ion.ParseTimestamp(t.Format("2006-01-02T15:04:05.000000Z"))`
* `timestamp := ion.NewTimestamp(t, ion.TimestampPrecisionNanosecond, ion.TimezoneUTC)`
* `timestamp := time.Date(2021, time.May, 25, 13, 41, 31, 90000, time.UTC)`

**Description of changes:**
Writing a timestamp nanosecond into Ion binary will write bytes of two values: Coefficient and exponent. These two values are used to calculate the nanoseconds using the equation: `coefficient * 10 ^ -exponent`.

For example a timestamp:  `2021-05-27 00:19:05.000050000` has coefficient: `50000` and exponent: `9`.

Inside `encodeTimeDate` function, we get the correct coefficient by calling `time.Nanosecond` which returns `50000` however the exponent is calculated by converting the coefficient to string and getting the length which returns `5` and is incorrect. The fix is to set the exponent to `9` if there are nanoseconds because `time.time` will always contain 9 precision units.

The next step is to read the binary ion timestamps correctly. Original implementation was to look at the coefficient value and remove trailing zeros but the binary ion already contains a byte containing the exponent so we can just use that.

Has gone through internal BQ review with comments in a [PR](https://github.com/byronlin13/ion-go/pull/1) to my fork.


ion/timestamp.go
```diff
@@ -471,23 +459,38 @@ func (ts Timestamp) String() string {
...
+		if timeZoneIndex != -1 && tIndex < timeZoneIndex {
 			zeros := strings.Builder{}
-			zeros.WriteByte('.')
-			for i := uint8(0); i < ts.numFractionalSeconds; i++ {
+			numZerosNeeded := 0
+
+			// Specify trailing zeros if fractional precision is less than the nanoseconds.
+			// e.g. A timestamp: 2021-05-25T13:41:31.00001234 with fractional precision: 2 will return "2021-05-25 13:41:31.00"
+			ns := ts.dateTime.Nanosecond()
+			if ts.dateTime.Nanosecond() == 0 || maxFractionalPrecision-len(strconv.Itoa(ns)) >= int(ts.numFractionalSeconds) {
```
justing-bq, line 478: `if ns == 0`

ion/timestamp.go
```diff
@@ -471,23 +459,38 @@ func (ts Timestamp) String() string {
...
+				zeros.WriteByte('.')
+				numZerosNeeded = int(ts.numFractionalSeconds)
+			} else {
+				dotIndex := strings.LastIndex(format, ".")
```
justing-bq, line 482: `decimalPlaceIndex`

ion/timestamp.go
```diff
@@ -471,23 +459,38 @@ func (ts Timestamp) String() string {
...
+				if dotIndex != -1 {
+					numZerosNeeded = int(ts.numFractionalSeconds) - (timeZoneIndex - dotIndex) + 1
```
justing-bq, line 484:
```
decimalPlaceIndex := strings.LastIndex(format, ".")
if decimalPlaceIndex != -1 {
	decimalPlacesOccupied := timeZoneIndex - decimalPlaceIndex - 1
	numZerosNeeded = int(ts.numFractionalSeconds) - decimalPlacesOccupied
}
```

ion/timestamp.go
```diff
@@ -159,30 +161,16 @@ func NewTimestampFromStr(dateStr string, precision TimestampPrecision, kind Time
 	if precision >= TimestampPrecisionNanosecond {
 		pointIdx := strings.LastIndex(dateStr, ".")
 		if pointIdx != -1 {
-			nonZeroFraction := false
```
justing-bq, line 164: Nit: let's get rid of this blank line.

byronlin13: Removed

ion/timestamp.go
byronlin13: Good catch, done.

ion/timestamp.go
byronlin13: Updated.

ion/timestamp.go
byronlin13: Updated.

ion/timestamp_test.go
```diff
@@ -208,3 +225,148 @@ func TestTimestampString(t *testing.T) {
...
+			kind := TimezoneUnspecified
+			if tt.fields.precision >= TimestampPrecisionMinute {
+				kind = TimezoneUTC
+			}
+
+			ts := &Timestamp{
+				dateTime:             dateTime,
+				precision:            tt.fields.precision,
+				kind:                 kind,
+				numFractionalSeconds: tt.fields.numFractionalSeconds,
```
zslayton, line 367: Am I missing something, or should this field have a name like `fractionalSecondsDecimalPlaces`? (i.e. It's a number of decimal places, not a number of fractional seconds.)

byronlin13: That is correct.
* `timestamp, err := ParseTimestamp("2006-01-02T15:04:05.123Z")` will have timestamp.numFractionalSeconds = 3
* `timestamp, err := ParseTimestamp("2006-01-02T15:04:05.123000Z")` will have timestamp.numFractionalSeconds = 6

ion/marshal.go
```diff
@@ -446,8 +445,9 @@ func (m *Encoder) encodeTimeDate(v reflect.Value) error {
 	// Get number of fractional seconds precisions
 	ns := t.Nanosecond()
 	numFractionalSeconds := 0
+	// Go native time object will always have 9 precision units when nanoseconds > 0
```
zslayton, line 448: Why does the precision change when `nanoseconds > 0`? Why isn't it a 0 of the same precision?

tgregg (commented):
Overall looks good, one comment and one question below.

ion/marshal.go
byronlin13: For `time.time` objects, if the nanosecond is 0 or not specified, it is treated as no nanoseconds.

`timeObject, err := time.Parse(time.RFC3339, "2000-01-01T01:01:01.00Z")` will have timeObject.nanoseconds = 0

`timeObject, err := time.Parse(time.RFC3339, "2000-01-01T01:01:01.01Z")` will have timeObject.nanoseconds = 10000000

zslayton: The [`Time` type documentation](https://golang.org/pkg/time/#Time) says:

> A Time represents an instant in time with nanosecond precision.

Unlike Ion's `Timestamp` type, which supports variable precision, it seems like `time.Time` _always_ has nanosecond precision. If that's true, I would expect a `time.Time` with zero nanoseconds to be written out as a `Timestamp` with a fractional seconds precision of 9 decimal places.

> For time.time objects, if the nanosecond is 0 or not specified, it is treated as no nanoseconds.

Could you please point me to the relevant docs for this? How do we distinguish between `0` or `not specified` in a `time.Time`?

ion/timestamp.go
```diff
@@ -515,12 +518,13 @@ func (ts Timestamp) Equal(ts1 Timestamp) bool {
 		ts.numFractionalSeconds == ts1.numFractionalSeconds
 }
 
-// TruncatedNanoseconds returns nanoseconds with trailing zeros removed (ie. 123456000 gets truncated to 123456).
+// TruncatedNanoseconds returns nanoseconds with trailing values removed up to the difference of max fractional precision - time stamp's fractional precision
+// e.g. 123456000 with fractional precision: 3 will get truncated to 123.
 func (ts Timestamp) TruncatedNanoseconds() int {
```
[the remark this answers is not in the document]
byronlin13: This function is only used internally to return the correct coefficient given the precision and is then used to write the timestamp in binary.

For example:
* `123456000` nanoseconds
* `3` fractional precision

This function will return coefficient: `123`. We will then write in binary, exponent: `3` and coefficient: `123`

So it will become `123 * 10 ^-3 = 0.123`

vs

Not truncating the nanoseconds which will be:
`123456789 * 10 ^-3 = 123456.789`

ion/bitstream.go
```diff
@@ -663,18 +663,7 @@ func (b *bitstream) ReadTimestamp() (Timestamp, error) {
...
-			for ns > 0 && (ns%10) == 0 {
-				ns /= 10
-				fractionPrecision--
-			}
-			precision = TimestampPrecisionNanosecond
-		} else if len(fracSecsBytes) > 0 && fracSecsBytes[0] > 0xC0 && (fracSecsBytes[0]^0xC0) > 0 {
+		if len(fracSecsBytes) > 0 && fracSecsBytes[0] > 0xC0 && (fracSecsBytes[0]^0xC0) > 0 {
 			fractionPrecision = fracSecsBytes[0] ^ 0xC0
 			precision = TimestampPrecisionNanosecond
 		}
```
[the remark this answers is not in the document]
byronlin13: We are getting the fraction precision using VarInt parsing within `b.readNsecs()` -> `b.readDecimal()` so I plumbed it through.

ion/marshal.go
byronlin13: You are correct, I was basing my interpretation on the fact that time.nanoseconds = 0 and the string representation was not showing fractional seconds 🤦 .

I removed the `nanoseconds > 0` check in  `encodeTimeDate()`.

I also added guards to  `NewDateTimestamp()`, `NewTimestamp()` and set `fractionalPrecision = 9` if the timestamp precision is `TimestampPrecisionNanosecond`

tgregg (approved):

ion/bitstream.go
```diff
@@ -692,32 +674,34 @@ func (b *bitstream) ReadTimestamp() (Timestamp, error) {
...
 	}
 
 	nsec, err = d.ShiftL(9).round()
 	if err != nil {
 		msg := fmt.Sprintf("invalid timestamp fraction: %v", d)
-		return 0, false, &SyntaxError{msg, b.pos}
+		return 0, false, 0, &SyntaxError{msg, b.pos}
 	}
 
+	exponent := uint8(d.scale)
```
justing-bq, line 697: Let's add a comment explaining that this cast should be safe because the max precision of 9 can fit inside uint8.

ion/timestamp_test.go
zslayton: Thanks for clarifying. I find this naming confusing, but it's part of the public API so we can't change it without a major version bump.

zslayton (approved):

ion/bitstream.go
byronlin13: That is not true because the [Ion timestamp spec](https://amzn.github.io/ion-docs/docs/spec.html#timestamp) mentions:

> Fractional seconds are allowed, with at least one digit of precision and an unlimited maximum.

Other Ion implementations can support arbitrary precision so Ion binary generated from Ion Java can have precision reduced.

Currently Ion Go support timestamps up to [9 precision units ](https://github.com/amzn/ion-go/releases/tag/v0.1.0-beta).

I created a [ticket](https://github.com/amzn/ion-go/issues/178) to update timestamps to support unlimited precision.

ion/binarywriter_test.go
```diff
@@ -186,7 +186,7 @@ func TestWriteBinaryTimestamp(t *testing.T) {
 	nowish, _ := NewTimestampFromStr("2019-08-04T18:15:43.863494+10:00", TimestampPrecisionNanosecond, TimezoneLocal)
 
 	testBinaryWriter(t, eval, func(w Writer) {
-		assert.NoError(t, w.WriteTimestamp(NewTimestamp(time.Time{}, TimestampPrecisionNanosecond, TimezoneUTC)))
+		assert.NoError(t, w.WriteTimestamp(NewTimestamp(time.Time{}, TimestampPrecisionSecond, TimezoneUTC)))
```
justing-bq, line 189: Why this change?

byronlin13: Because the [binary data](https://github.com/amzn/ion-go/pull/177/files/b3422b0e231c904ff2846e832ba54a374bbd42ea#diff-a9e03e3f1ff004ab18b39215530677a79ee3745174b52160e5d525d62240b7e1R173) doesn't have nanoseconds. It was failing because calling [NewTimestamp](https://github.com/amzn/ion-go/pull/177/files/b3422b0e231c904ff2846e832ba54a374bbd42ea#diff-4bca5f01731890e285690a1055a0f4e75e03539170cfd8f3fe6c57d6e2918b78R150) with TimestampPrecisionNanosecond will set fractional precision: 9.

Issue #184, opened 2021-09-29
Unmarshaler interface


lindlof:
The `Marshaler` interface is provided for marshaling custom types to ion. However, no interface is provided for unmarshaling from ion. Is there some technical reason to not provide this?

It would be immensely helpful for building types which are interoperable with both ion and json. Specifically, we're looking to have decimal and timestamp types which are interoperable with both ion and json.

fernomac:
As far as I remember (it's been a while), no I just never got around to it. 😅 Taking a quick look, I think it _should_ just be a matter of defining the interface and dropping something like [this](https://github.com/amzn/ion-go/blob/3f5d6c7f1616ca404a136803c1b08405d7a29b31/ion/marshal.go#L202) in [here](https://github.com/amzn/ion-go/blob/master/ion/unmarshal.go#L307)?

errorhandler:
Just jumping into 👍🏻   the need for this, I've thrown together a quick branch [here](https://github.com/JasperLabs/ion-go/tree/unmarshal) that I'm happy to contribute back!

desaikd:
Thanks for putting together a quick branch @errorhandler! Can you please create a pull request for this? It would be easier for us to review it over there.

popematt:
I'm going to close this now that #185 has been merged. Thanks to @errorhandler for implementing this!

PR #185, opened 2022-01-31
#184: Allow custom unmarshal implementation

errorhandler:
Issue #, if available: #184

Description of changes:
Adds the unmarshalling equivalent of `Marshaler`. Allowing types to implement their own unmarshalling logic. It's basically the same implementation, but requires the `UnmarshalIon` method to be implemented with a pointer receiver as it needs to mutate the struct.

### DIFF
--- a/ion/unmarshal.go
+++ b/ion/unmarshal.go
@@ -38,6 +38,11 @@ var typesAcceptableKinds = map[Type][]reflect.Kind{
 	ListType:  {reflect.Slice, reflect.Array},
 }
 
+// Unmarshaler is the interface implemented by types that can unmarshal themselves to Ion.
+type Unmarshaler interface {
+	UnmarshalIon(r Reader) error
+}
+
 // Unmarshal unmarshals Ion data to the given object.
 //
 // User must pass the proper object type to the unmarshalled Ion data.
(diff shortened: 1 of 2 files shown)


popematt:
Hi, thanks for your contribution!

It looks like the CI build failed because one of our Github Workflows was broken. I've fixed the workflow now. Do you mind rebasing your commit against the tip of `master` in `amzn/ion-go` so that we can re-run with the fixed workflow? Alternately, you could grant me permission to edit the PR, and I'll update it.

ion/unmarshal.go
```diff
@@ -38,6 +38,11 @@ var typesAcceptableKinds = map[Type][]reflect.Kind{
 	ListType:  {reflect.Slice, reflect.Array},
 }
 
+// Unmarshaler is the interface implemented by types that can unmarshal themselves to Ion.
```
popematt, line 41:
```suggestion
// Unmarshaler is the interface implemented by types that can unmarshal themselves from Ion.
```

popematt (changes requested):
Aside from one very small typo, this looks good. Thank you!

popematt (approved):

Vingtoft:
HI, can you please provide an example of how to use UnmarshalIon?

Given the struct:

```
type Account struct {
	Name         string          `ion:"name"`
}
```

I write the function:

```
func (a *Account) UnmarshalION(b []byte) error {
    fmt.Println("In Unmarshall")
    a.Name = "boo"
    return nil
}
```

The above implementation does not work.

Needless to say: Im learning Go. 

Thanks!

PR #194, opened 2024-04-25
Fix math/big.Int Does not Marshal

jeffmitchell-sas:
This is intended to be a fix for https://github.com/amazon-ion/ion-go/issues/193. ion.MarshalText() returns "{}" for math/big.Int values.

Added BigIntType in type.go. Created func encodeBigInt() in marshal.go and added it to encodeStruct(). Also added unit tests for math/big.Int in marshal_test.go.

### DIFF
--- a/ion/marshal.go
+++ b/ion/marshal.go
@@ -384,6 +384,9 @@ func (m *Encoder) encodeStruct(v reflect.Value) error {
 	if t == decimalType {
 		return m.encodeDecimal(v)
 	}
+	if t == bigIntType {
+		return m.encodeBigInt(v)
+	}
 
 	if err := m.w.BeginStruct(); err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


popematt:
Thanks for contributing this!

It looks like your change will make `big.Int` marshal correctly, but I'm surprised that you needed to make changes in `ion/type.go`. We don't want to add a new BigInt type to Ion because in Ion, a "big" int is still an int. Does the marshaling still work if the changes to that file are reverted? TBH, I haven't personally touched this code in quite a while, so I don't actually know if there's something that I'm missing here. I'll need to look into it more.

jeffmitchell-sas:
Good call. I removed the changes in type.go and it still marshalled correctly. I misinterpreted types.go. I reverted the changes to that file. Thanks!

popematt (approved):

nirosys (approved):
LGTM. Thank you Jeff!

jeffmitchell-sas:
Thanks guys!

jeffmitchell-sas:
What is the release schedule like? When is the next release normally tagged?

tgregg:
> What is the release schedule like? When is the next release normally tagged?

I will create a new release by Monday.

tgregg:
@jeffmitchell-sas Released in https://github.com/amazon-ion/ion-go/releases/tag/v1.4.0

jeffmitchell-sas:
Thank you @tgregg

PR #196, opened 2024-06-03
Add Reset function to binaryReader

jt0:
Reset improves reader performance when the same binary data is processed multiple times. It works by reusing the symbol table constructed during the first read, and jumping straight to the first "content" byte when Reset is called.

### DIFF
--- a/ion/binaryreader.go
+++ b/ion/binaryreader.go
@@ -24,8 +24,9 @@ import (
 type binaryReader struct {
 	reader
 
-	bits bitstream
-	cat  Catalog
+	bits     bitstream
+	cat      Catalog
+	resetPos uint64
 }
 
 func newBinaryReaderBuf(in *bufio.Reader, cat Catalog) Reader {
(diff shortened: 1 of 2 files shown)


ion/binaryreader.go
```diff
@@ -36,6 +37,36 @@ func newBinaryReaderBuf(in *bufio.Reader, cat Catalog) Reader {
...
+//
+// While it is possible to call Reset with different input bytes, the Reader
+// will only work correctly if the new bytes contain the exact same binary
+// version marker and local symbols as the original input. If there are any
+// doubts whether this is the case, it is instead recommended to create a
+// new Reader using NewReaderBytes (or NewReaderCat) instead. Attempting to
+// reuse a binaryReader with inconsistent input bytes will cause the reader
+// to return errors, misappropriate values to unrelated or non-existent
+// attributes, or incorrectly parse data values.
+func (r *binaryReader) Reset(in []byte) error {
```
popematt, line 54: If you don't mind, can you add a note about the visibility of this function for those of us who aren't using Go on a regular basis?

popematt:
@nirosys, Can you take a look at this PR?

nirosys (commented):
Hey jt0! Thanks for the PR.

Could you maybe provide some more information about your use case, and what prompted you to look into this optimization?

I have some concerns about the approach, but I might be missing something.

ion/binaryreader.go
jt0: The function is exported, so anyone can call it. It's not included in the `Reader` interface, which is the return type from the `NewReader*` functions. In order to use this, one needs to define their own Go interface similar to this:

```go
type resettableReader interface {
	ion.Reader
	Reset([]byte) error
}
```

and perform a type assertion on the interface in order to use it. For example:

```go
func resetOrNew(reader ion.Reader, content []byte) ion.Reader { 
	if rr, ok := reader.(resettableReader); ok {
		_ = rr.Reset(content)
		return rr
	}
	return ion.NewReaderBytes(content)
}
```

nirosys (dismissed):
After some offline discussion, I'm good with merging the change, with the understanding that we'll revisit the use case at a later date.

ion/binaryreader.go
```diff
@@ -36,6 +37,36 @@ func newBinaryReaderBuf(in *bufio.Reader, cat Catalog) Reader {
...
+// when the same data needs to be read multiple times.
+//
+// While it is possible to call Reset with different input bytes, the Reader
+// will only work correctly if the new bytes contain the exact same binary
+// version marker and local symbols as the original input. If there are any
+// doubts whether this is the case, it is instead recommended to create a
+// new Reader using NewReaderBytes (or NewReaderCat) instead. Attempting to
+// reuse a binaryReader with inconsistent input bytes will cause the reader
+// to return errors, misappropriate values to unrelated or non-existent
+// attributes, or incorrectly parse data values.
```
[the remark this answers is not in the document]
jt0: > In ion-c we have APIs to reset the underlying stream of a reader, which includes resetting the catalog and symbol table to default values.

In ion-go, you achieve this semantic by calling `ion.NewReaderBytes` or `ion.NewReader` with an io.Reader that has it's own reset capability.

> That it doesn't work "correctly" unless you provide bytes that have the exact same leading bytes also concerns me. I feel like the method would be very error prone

I tried to be clear that one has to be careful to only provide consistent bytes. Tbh, I don't expect that would be a common use case. Happy to update the comment if it needs to be more explicit.

> which leads me to think that if the application had the right information (catalog, and reader offset), it could re-create the reader itself.

Not as the library is currently written, since neither binaryReader's nor its attributes are exported, and there are no "Set" functions provided. Even if those were added, I'm not sure it would be any better than having a function like Reset that manages the details for you.

nirosys (approved):

